PR #1, opened 2016-12-10
Add ts deps to default blueprint

dustinfarris:
Simplify install:

```
ember install "ember-cli-typescript@emberwatch/ember-cli-typescript"
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -3,17 +3,12 @@
 Enable typescript preprocessing on Ember 2.x apps.
 
 
-## One time setup
+## Installation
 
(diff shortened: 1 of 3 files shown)


chriskrycho:
Note that as a user of this, I'd appreciate a note in the README that explained that this would happen by default, and how to opt out if you wanted a different strategy. I actually like the inclusion, but it's worth making explicit precisely because of the way transitive dependency issues can wreak havoc.

dustinfarris:
@chriskrycho yup, good point

PR #2, opened 2016-12-10
Further clarification of ember install effects

dustinfarris:
/cc @chriskrycho please review

### DIFF
--- a/README.md
+++ b/README.md
@@ -5,9 +5,13 @@ Enable typescript preprocessing on Ember 2.x apps.
 
 ## Installation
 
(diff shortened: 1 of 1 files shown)


dustinfarris:
Will probably need to include a note for addon authors as well since ember puts these in devDependencies which probably won't work for addons.

winding-lines:
Dustin, are you testing your changes in a typescript application? 

Marius
---

dustinfarris:
@winding-lines yes I installed successfully using `ember install`, however note that deps are added to `devDependencies` not `dependencies` which I believe still needs to be clarified for addon authors.

dustinfarris:
@winding-lines to be clear i only tested that the install process had the same effect

- same packages installed
- package.json updated
- typescript runs during build without errors

I have not tested that my addon works with ember-cli-typescript in general.  I am doing that now.

PR #6, opened 2016-12-29
README.md: Fixed typo

perlun:
Also changed from the weasel-sounding "better" to "greater", which is more clear and to the point IMHO.


locks:
Appreciated!

PR #20, opened 2017-04-15
Fix general install issues

chriskrycho:
Things I've tried which *don't* resolve the issue:

- using the `beforeInstall` hook instead of the `afterInstall` hook

### DIFF
--- a/blueprints/ember-cli-typescript/index.js
+++ b/blueprints/ember-cli-typescript/index.js
@@ -24,7 +24,7 @@ module.exports = {
     // Entity name is optional right now, creating this hook avoids an error.
   },
 
-  beforeInstall: function() {
+  afterInstall: function() {
     return this.addPackagesToProject([
       { name: 'typescript', target: '^2.1' },
       { name: '@types/ember', target: '^2.7.34' }
--- a/blueprints/ember-cli-typescript/index.js
+++ b/blueprints/ember-cli-typescript/index.js
@@ -24,7 +24,7 @@ module.exports = {
     // Entity name is optional right now, creating this hook avoids an error.
   },
 
-  afterInstall: function() {
+  beforeInstall: function() {
     return this.addPackagesToProject([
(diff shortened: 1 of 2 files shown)


pixelhandler:
@chriskrycho per our slack chat just wanted to note your observation…

> **I think broccoli-typescript-compiler should make typescript a dependency.**

chriskrycho:
Good add, @pixelhandler.

To elaborate on that: `index.js` requires `lib/typescript-preprocessor.js`, and the `TypeScriptPreprocessor` instance there gets instantiated during the add-on installation process. That explicitly depends on both `typescript` and `broccoli-typescript-compiler`—the latter of which is installed, but the former of which is not, because none of the packages, including `broccoli-typescript-compiler`, specify it in their `dependencies`—it's in `peerDependencies` in `broccoli-typescript-compiler`.

chriskrycho:
Per @Turbo87, I just went with an approach that lazily requires the dependencies, so that when they're missing during install, it just skips the steps that depend on it. My initial testing works fine.

@pixelhandler @locks can someone do a clean install using this branch on a *nix and confirm that it's working? Because it's on a branch, that should just be:

```
$ ember install git+ssh://git@github.com:chriskrycho/ember-cli-typescript.git#before-install
$ ember generate ember-cli-typescript
```

That should complete successfully (which verifies this fixes the issue it set out to fix); and you should be able to generate a normal item (`ember g component something-typescript` or whatever), change the file extension to `.ts`, and have it work as expected.

pixelhandler:
@chriskrycho yeah just tried those two commands on my mac. Works like a charm :)

I also ran builds for all envs (`--prod` and `--env test`) and ran server as well as `ember t -s`, seems to run fine in a new app after install.

pixelhandler:
@locks can you review / merge ?

pixelhandler:
@locks thanks 👍 perhaps release 0.4.0 next week :)

PR #22, opened 2017-04-19
Set correct default module resolution in tsconfig.

### LINKED ISSUE #4: Current blueprint does not work for *editor* module resolution
In order to get module resolution working properly in VS Code—i.e., so that code completion and navigation-to-definition worked—I had to make the following changes to the blueprint. Note the additions to the `paths` and `include` keys.

```json
{
  "compilerOptions": {
    "target": "ES6",
    "allowJs": true,
    "moduleResolution": "node",
    "noEmitOnError": true,
    "noEmit": true,
    "baseUrl": ".",
    "paths": {
      "mobile-web/config/*": ["config/*"],
      "mobile-web/tests/*": ["tests/*"],
      "mobile-web/*": ["app/*"],
      "npm:*": ["local-types/*"],
      "*": ["local-types/*"]
    }
  },
  "include": [
    "app/**/*",
    "tests/**/*"
  ]
}
```

If possible, these should either be generated by default, or the README should include them (I'm happy to submit a PR for either change, but wanted to open for discussion). It may also be worth bringing up the specific issue highlighted in the [Configuration file](https://github.com/emberwatch/ember-cli-typescript/blob/9cae86bd36f95174953b8f714f70397e91ea16c3/README.md#configuration-file) section with the TypeScript team, as it seems like it may well affect more than just us.

(I'm concerned from a comment @locks made in Slack that this may end up including the `tests` directory during the processing, but that seems like a decision that should be made _during_ the Broccoli filtering, just like it would be in the pure-JS case, right?)

### LINKED ISSUE #21: tsconfig questions/issues
- What is `"local-types/*"` ([add-on][a-local-types], [blueprint][b-local-types])? I find no local reference to this and nothing documenting anything like it for TS in general.
- What is up with [`"node-modules/at-types-ember"`][at-types-ember]? Shouldn't this just be `"node-modules/@types/ember"`?

[a-local-types]: https://github.com/emberwatch/ember-cli-typescript/blob/6db76ca1a170d84a665b60c3e5a30d601cfa558b/tsconfig.json#L10-L11
[b-local-types]: https://github.com/emberwatch/ember-cli-typescript/blob/6db76ca1a170d84a665b60c3e5a30d601cfa558b/blueprints/ember-cli-typescript/files/tsconfig.json#L10-L11
[at-types-ember]: https://github.com/emberwatch/ember-cli-typescript/blob/6db76ca1a170d84a665b60c3e5a30d601cfa558b/tsconfig.json#L9

chriskrycho:
Fixes #4, closes #21.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -19,7 +19,3 @@ yarn.lock
 
 # TypeScript
 jsconfig.json
-
-# JetBrains
-.idea
-*.iml
--- a/.gitignore
+++ b/.gitignore
@@ -19,3 +19,7 @@ yarn.lock
 
 # TypeScript
 jsconfig.json
+
+# JetBrains
+.idea
+*.iml
--- a/tsconfig.json
+++ b/tsconfig.json
@@ -6,9 +6,7 @@
     "noEmitOnError": true,
     "baseUrl": ".",
     "paths": {
-      "ember": ["node_modules/at-types-ember"],
-      "npm:*": ["local-types/*"],
-      "*": ["local-types/*"]
+      // @types all included implicitly
     }
   },
   "include": [
--- a/tsconfig.json
+++ b/tsconfig.json
@@ -5,9 +5,7 @@
     "moduleResolution": "node",
     "noEmitOnError": true,
     "baseUrl": ".",
-    "paths": {
-      // @types all included implicitly
-    }
+    "paths": {}
   },
   "include": [
     "app/**/*"
--- a/blueprints/ember-cli-typescript/files/tsconfig.json
+++ b/blueprints/ember-cli-typescript/files/tsconfig.json
@@ -3,15 +3,16 @@
     "target": "ES6",
     "allowJs": true,
     "moduleResolution": "node",
-    "noEmitOnError": true,
+    "noEmitOnError": false,
     "noEmit": true,
     "baseUrl": ".",
     "paths": {
-      "npm:*": ["local-types/*"],
-      "*": ["local-types/*"]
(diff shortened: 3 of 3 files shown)


tsconfig.json
```diff
@@ -6,9 +6,7 @@
     "noEmitOnError": true,
     "baseUrl": ".",
     "paths": {
-      "ember": ["node_modules/at-types-ember"],
-      "npm:*": ["local-types/*"],
-      "*": ["local-types/*"]
+      // @types all included implicitly
```
poteto: This is invalid json

chriskrycho: @poteto TRUE! This is what happens when you leave a note for yourself. Thanks for catching it.

.gitignore
```diff
@@ -19,3 +19,7 @@ yarn.lock
 
 # TypeScript
 jsconfig.json
+
+# JetBrains
```
chriskrycho: I don't ignore it in every project; it's quite helpful to me for projects where I actually want the JetBrains IDE files (usually not ones I'm sharing), which is why I don't have it in global. I *can* ignore it at that level, but as soon as someone *else* adds it and doesn't have that… ¯\\\_(ツ)_/¯ I'm a fan of "ignore everything like this instead of accidentally committing it and being sad"; but that's me.

locks: Sadness is unavoidable, Chris.

chriskrycho: Found a better solution: `.git/info/exclude`: per-user gitignores. (Who knew? Not me till two minutes ago.)

PR #26, opened 2017-04-24
Add additional config declarations

luketheobscure:
Using default app.js from ember-cli generator declares these properties, which raise warnings in Typescript.

I'm brand new to typescript, so I'm not really sure this is the way to solve this, so take this PR with a grain of salt.

### DIFF
--- a/blueprints/ember-cli-typescript/files/app/config/environment.d.ts
+++ b/blueprints/ember-cli-typescript/files/app/config/environment.d.ts
@@ -9,4 +9,7 @@ export default config;
  */
 declare namespace config {
   export var environment: any;
+  export var modulePrefix: String;
+  export var podModulePrefix: String;
+  export var locationType: String;
 }


chriskrycho:
Awesome, @luketheobscure! I was hoping someone would get to that!

blueprints/ember-cli-typescript/files/app/config/environment.d.ts
```diff
@@ -9,4 +9,7 @@ export default config;
  */
 declare namespace config {
   export var environment: any;
+  export var modulePrefix: String;
```
leepfrog: `String` should be `string`  -- i'm new to typescript and made this mistake :)

https://www.typescriptlang.org/docs/handbook/declaration-files/do-s-and-don-ts.html

leepfrog: @luketheobscure @chriskrycho

luketheobscure: Thanks for the heads up. Also brand new to ts.

https://github.com/emberwatch/ember-cli-typescript/pull/27

PR #32, opened 2017-05-01
Update README with latest changes.

chriskrycho:
This doesn't change anything functionally; it just makes it a tad more modern and maintainable. (No need to do a new release, accordingly.)


pixelhandler:
@chriskrycho it's really helpful to have the "not yet" supported section in the README. Likewise the instructions for addon use is great, I had no idea that's why I had to use only 'app' directory with an addon.

PR #40, opened 2017-06-04
[FEATURE]: support non @types based TS type defs

toranb:
This is half observation / half question. I've found working outside of ember that most library authors now ship the `index.d.ts` file w/ the JS source (instead of using the definitely typed repository). This change does incur a tax today because we don't filter that types work down to only `**/*.d.ts` file types so ... in my app I had to hack a ton of modules out using exclude (not ideal)

Any chance we could expand this type defs lookup to reach the wider JS ecosystem by not zooming in on the `node_modules/@types` dir explicitly?

```js
  "exclude": [
      "node_modules/typescript/lib/**/*.ts",
      "node_modules/memory-streams/**/*.ts",
      "node_modules/ember-source/**/*.ts",
      "node_modules/@glimmer/**/*.ts",
      "node_modules/ember-cli-typescript/**/*.ts",
      "node_modules/iconv-lite/lib/**/*.ts",
      "node_modules/aot-test-generators/**/*.ts",
      "node_modules/rx/ts/**/*.ts"
  ]
```

### DIFF
--- a/lib/typescript-preprocessor.js
+++ b/lib/typescript-preprocessor.js
@@ -22,7 +22,7 @@ function readConfig(configFile) {
  * Return the paths which contain type information.
  */
 function typePaths(config) {
-  const base = ["node_modules/@types"];
+  const base = ["node_modules"];
 
   const cfgPaths = (config.compilerOptions && config.compilerOptions.paths) || {};
 


rwjblue:
In general, we should let `tsc` do its job and resolve these types for us.  The more we do to "get in the middle", the more headache we will have 😺 .  A recent refactor of broccoli-typescript-compiler enables us to stop passing down the `node_modules` tree into typescript, and instead rely on `tsc`'s normal resolution system. I'm hopeful that @krisselden will be able to release at least a alpha/beta version soon of that so we can start testing here...

toranb:
@rwjblue ah I didn't realize core was aware of this issue. So what is the correct location of these type def files so typescript can resolve them w/out any special configuration? At one time it seemed like you would put them in @types/your-lib but today I see a lot of type defs shipping with the source code (to match version/ etc).

toranb:
For anyone else who might follow - I found a better workaround than hacking the source/forking this addon. If you have a few typedefs that `ember-cli-typescript` cannot find, crack open the tsconfig and add a few one off paths like this

```js
{
  "compilerOptions": {
    "target": "es2015",
    "module": "es2015",
    "moduleResolution": "node",
    "strictNullChecks": true,
    "noImplicitAny": true,
    "paths": {
      "welp/*": ["app/*"],
      "ember-redux": ["node_modules/ember-redux/index.d.ts"],
      "redux": ["node_modules/redux/index.d.ts"]
    }
  },
  "include": [
    "**/*.ts"
  ]
}
```

chriskrycho:
Yep, and this is the preferred way of resolving files in general with TypeScript. See the fuller discussion in the TypeScript docs [here](http://www.typescriptlang.org/docs/handbook/module-resolution.html#path-mapping).

@toranb if you have a few minutes and want to add a note to the README along those lines I'd merge it in a heartbeat.

PR #53, opened 2017-08-02
fix error transpiling referenced javascript files

dwickern:
When referencing a .js from .ts:
```
  // service.ts
  import User from './user.js';
```
This error will occur:
```
  error TS5055: Cannot write file './user.js' because it would overwrite input file.
```


chriskrycho:
🎉 Thanks so much!

PR #54, opened 2017-08-02
add in-repo-addons to generated tsconfig.json

dwickern:
Adds path for in-repo-addons / engines, for example:
```json
{
  "compilerOptions": {
    ...
    "paths": {
      "web/tests/*": ["tests/*"],
      "web/*": [
        "app/*",
        "lib/in-repo-addon-1/app/*",
        "lib/in-repo-addon-2/app/*"
      ]
    }
  }
}
```

I also added some tests for it

### DIFF
--- a/.npmignore
+++ b/.npmignore
@@ -14,3 +14,4 @@
 bower.json
 ember-cli-build.js
 testem.js
+node-tests/
--- a/.travis.yml
+++ b/.travis.yml
@@ -35,6 +35,7 @@ script:
   # Usually, it's ok to finish the test scenario without reverting
   #  to the addon's original dependency state, skipping "cleanup".
   - node_modules/.bin/ember try:one $EMBER_TRY_SCENARIO test --skip-cleanup
+  - yarn run nodetest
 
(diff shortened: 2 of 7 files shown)


chriskrycho:
1. Awesome!

2. *Doubly* awesome because you added tests for it. 🎉 

3. Does it actually need the specific names of the in-repo add-ons? I *think* it would work if you just did this:

    ```json
    {
      "compilerOptions": {
        "paths": {
          "web/tests/*": ["tests/*"],
          "web/*": [
            "app/*",
            "lib/*"
          ]
        }
      }
    }
    ```

4. Can you merge `master` in (*not* rebase; I prefer a readable history in this case).

dwickern:
You do need the whole path to the `app` directory because imports are relative to that directory. For example,

`import MyComponent from 'my-app/components/my-component'`

would normally resolve to
`app/components/my-component.js`

 but can also resolve to 
`lib/<in-repo-addon-name>/app/components/my-component.js`

chriskrycho:
Ah, that makes good sense. I'll merge it once Appveyor passes.

PR #59, opened 2017-08-16
Generate ES6 modules (let ember-cli-babel do its thing).

chriskrycho:
This resolves a problem where tools that codemod the imports,
e.g. babel-plugin-ember-modules-api-polyfill, were unable to
work because the modules being generated were not the same ES6
modules as the codemod expected (TS generates CommonJS modules
by default).


chriskrycho:
AppVeyor build [succeeded](https://ci.appveyor.com/project/chriskrycho/ember-cli-typescript). Was a mismatch with web hooks from when we moved the repo. Should be resolved for future builds, I expect.

PR #61, opened 2017-08-19
Use console-ui for TS diagnostics

simonihmig:
Upgrading to e-c-t 1.0 broke our CI setup, because TS warnings were not emitted to stderr (`console.error`) anymore, so went to our JUnit xml file which obviously broke its format, thus the CI server could not parse it anymore. Looks like this was changed in `broccoli-typescript-compiler` on 2.0: https://github.com/tildeio/broccoli-typescript-compiler/commit/379adc05add513a9f56c3a6da7c0086f5b2d4a03#diff-d5edbe17e79bd7d0417b0bac7e92be06L87

Luckily there is an API to change the writer function, so we can use ember-cli's `ui` from ember-cli for all messages. This now also correctly supports the `--silent` flag to suppress any of those messages.

### DIFF
--- a/index.js
+++ b/index.js
@@ -24,7 +24,9 @@ module.exports = {
     }
 
     try {
-      registry.add('js', new TsPreprocessor());
+      registry.add('js', new TsPreprocessor({
+        ui: this.ui
+      }));
     } catch (ex) {
       this.ui.write(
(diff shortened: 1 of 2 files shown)


rwjblue:
Seems good to me...

chriskrycho:
Oh, nice catch—I had no idea! Thank you so much for fixing this.

PR #62, opened 2017-08-19
Throw SilentError when preprocessor could not be instantiated

simonihmig:
To further improve on message handling... I think it is reasonable to cancel the build processes when the TS preprocessor has not been initialized correctly, as the build output without it will certainly be broken. Also this gives a clear error indication (rendered in red color by ember-cli).

### DIFF
--- a/index.js
+++ b/index.js
@@ -1,26 +1,25 @@
 /* eslint-env node */
 
+const fs = require('fs');
+const path = require('path');
 const SilentError = require('silent-error');
+const TsPreprocessor = require('./lib/typescript-preprocessor');
 
-let TsPreprocessor;
-try {
-  TsPreprocessor = require('./lib/typescript-preprocessor');
-} catch (ex) {
-  // Do nothing; we just won't have the plugin available. This means that if you
-  // somehow end up in a state where it doesn't load, the preprocessor *will*
-  // fail, but this is necessary because the preprocessor depends on packages
-  // which aren't installed until the
-}
 
 module.exports = {
   name: 'ember-cli-typescript',
 
   setupPreprocessorRegistry(type, registry) {
-    if (!TsPreprocessor) {
-      this.ui.write(
-        'Note: TypeScript preprocessor not available -- some dependencies not installed. ' +
-          '(If this is during installation of the add-on, this is as expected. If it is ' +
-          'while building, serving, or testing the application, this is an error.)'
+    if (!fs.existsSync(path.join(this.project.root, 'tsconfig.json'))) {
+      // Do nothing; we just won't have the plugin available. This means that if you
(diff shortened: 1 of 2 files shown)


rwjblue:
Agreed.

chriskrycho:
This seems great, @simonihmig – can you take a look at the build failure and get that sorted? Super happy to have this change in place once that's cleared up!

chriskrycho:
I'll add: as soon as I see that this is passing, I'll not only merge but publish as 1.0.3!

simonihmig:
@chriskrycho Oh, good that there were these tests, the changes would indeed have caused some problems otherwise! 😬

But I think the error handling was somewhat broken before: 
* [here](https://github.com/typed-ember/ember-cli-typescript/blob/62a6d8eda008d1f2e2ec2987c63db116c2eee015/index.js#L5) the `TsPreprocessor` constructor was just imported without running it, so the try/catch was not really doing anything
* given that, the guard [here](https://github.com/typed-ember/ember-cli-typescript/blob/62a6d8eda008d1f2e2ec2987c63db116c2eee015/index.js#L17-L24) would never been executed
* that's why the tests broke: when there is no `tsconfig.json` (like in the tests before running the blueprint), it would run into the line where we now throw the error. 

I changed this a bit to check if there is a `tsconfig.json` and do the warning then and skip the rest. Hope that makes sense, but would appreciate your 👀, @chriskrycho & @rwjblue!

chriskrycho:
Nice work. Quick question: have you tried a clean install and `ember generate ember-cli-typescript` from this branch? The original hooks were there to handle some funkiness in the install sequence. I think those have been ironed out with a bunch of the changes we’ve since made, but I hadn’t checked on those in a few months (and they preceded my time as a maintainer). If you can double-check that works, I’ll confirm it Tuesday and merge and publish. (Tuesday because I’m going to be off and traveling to see the solar eclipse tomorrow!)

simonihmig:
Sure, I just double checked. As I cannot just run `ember install ember-cli-typescript` with my branch, I did the following:
* in a fresh app I added e-c-t to package.json and `yarn link`ed it
* with a simple ember-cli command like `ember t` there is now the "Skipping ..." warning
* `ember g ember-cli-typescript` installed `tsconfig.json` and the required deps as expected
* `ember t` is all good, no warning messages anymore
* renamed `app.js` to `app.ts`
* `ember t` is still all good

So seems good to me!

simonihmig:
Don't know what hard time appveyor is having, seems stuck somewhere...

chriskrycho:
I think it probably passed; I literally just need to turn the integration off and on again (it got confused when when moved the repository for coordinating all things TS).

PR #69, opened 2017-09-13
Remove unnecessary export keyword in blueprint

dwickern:
I'm pretty sure `export` is implied. The typings seem to work without it.


chriskrycho:
The build is breaking because of the changes to RSVP. 😒 Also, why isn't Travis reporting status? Argh.

chriskrycho:
@dwickern even worse: after some investigation, the problem is that when the DefinitelyTyped modules are shipped, the tsconfig is apparently stripped. So having that setting local to Ember.js' typings and even in this repo fails. 😑 How far do you think we are from being able to do a release of the Ember typings you and @dfreeman have been working on for users to depend on?

dwickern:
So my understanding from DefinitelyTyped/DefinitelyTyped#18872 is that our ember typings need to `import RSVP = require('rsvp')` and our tsconfig blueprint needs to have `allowSyntheticDefaultImports`?

chriskrycho:
I *think* that *should* work, I *guess*? Ugh. Will allocate an hour to this tomorrow.

chriskrycho:
As with #83, the Travis failure is spurious. Underlying issue is https://github.com/ember-cli/ember-try/issues/161.

PR #71, opened 2017-09-14
enable `allowSyntheticDefaultImports` in tsconfig

dwickern:
This will make these imports work
```ts
import $ from 'jquery';
import RSVP from 'rsvp';
```

Since those typings use `export =` instead of `export default`


chriskrycho:
That's what I tried earlier, but no dice. 😖

dwickern:
The error now is 
> node_modules/@types/ember/index.d.ts(13,8): error TS1192: Module '"/home/travis/build/typed-ember/ember-cli-typescript/node_modules/@types/rsvp/index"' has no default export.

So I think we need to update the ember typings also (or pin `@types/rsvp` to an older version for now)

chriskrycho:
Ah, I see. Let's pin to the earlier version for now.

chriskrycho:
Here's a question I don't know the answer to: how do DT types work with real package dependencies? My goal is to upstream RSVP types to RSVP itself by the end of the week. I have no idea, though, what the answer is to this question.

dwickern:
They let you add a `package.json` with a `"dependencies"` (and nothing else). I think that's why

https://github.com/DefinitelyTyped/DefinitelyTyped#i-notice-some-packages-having-a-packagejson-here

dwickern:
You would also need to update the notNeededPackages on DT for RSVP

chriskrycho:
👍 excellent; thank you. I'd poked around at that question before but never found my way to that answer. I'll get that rolling probably Thursday.

chriskrycho:
I *think* this is now defunct, b/c we have RSVP@4.0.0+ in DT and that's _not_ using synthetic exports?

chriskrycho:
Same as other recently merged branches: Travis failure is spurious.

PR #76, opened 2017-10-05
Update README with comment on environment.d.ts

theroncross:
Had to track this down the other day when I couldn't figure out how there was a 'config' type at all.


chriskrycho:
(Will merge as soon as AppVeyor is green. Goodness but Node on Windows is slow.)

PR #84, opened 2017-11-23
Add noImplicitThis in default tsconfig

inkless:
This is related to https://github.com/Microsoft/TypeScript/issues/14518
When `noImplicitThis` is not on, `this` is not `ThisType<T>`, so we should turn on `noImplicitThis` by default

### DIFF
--- a/blueprints/ember-cli-typescript/files/tsconfig.json
+++ b/blueprints/ember-cli-typescript/files/tsconfig.json
@@ -4,6 +4,7 @@
     "allowJs": true,
     "moduleResolution": "node",
     "allowSyntheticDefaultImports": true,
+    "noImplicitThis": true,
     "noEmitOnError": false,
     "noEmit": true,
     "sourceMap": true,


chriskrycho:
Ohhh, interesting. I'll see how this behaves in my app, and will merge after confirming. Thank you for the contribution!

chriskrycho:
Curiously, this fixes it nicely for the `EmberObject.extend({ ... })` case but doesn't help *at all* with ES6 classes. It's a nice win for the non-class variant, though, so I'll take it gladly. Thanks again!

dwickern:
What's wrong with ES6 classes?

chriskrycho:
@dwickern without using a `this` binding, you end up with errors like this:

error

The only way around it I've found so far is explicitly naming `this`:

```ts
import Component from '@ember/component';

export default class MyFoo extends Component {
  thingToSay = 'Hello';

  sayIt(this: MyFoo, toSomeone: string) {
    const thingToSay = this.get('thingToSay'); // errors without `this: MyFoo` above
    alert(`${thingToSay} ${toSomeone}`);
  }
}  
```

dwickern:
Are you using `strictFunctionTypes`?

chriskrycho:
I had no been, but turning it on has no effect on this error, either.

PR #85, opened 2017-11-24
Implement fast incremental compiler

dwickern:
The idea here is to use `tsc --watch` instead of broccoli-typescript-compiler, since `tsc` has much faster incremental compile logic. There is no public API for the incremental compiler yet, but once Microsoft/TypeScript#9282 is implemented, broccoli-typescript-compiler should be equally fast and this PR will be obsolete.

There are lots of issues with my hacked together implementation
- the `.ts` are still being watched, so each change triggers _two_ rebuilds. Fixing this should make it faster.
- ember-cli will not watch a directory in `tmp` so I put the compiled js into an `.e-c-ts` directory
- the incremental compiler should only be used in development, but I couldn't figure out how to check the environment
- I couldn't figure out the broccoli hook to use to clean up temp files and child processes
- I'm swallowing all the stdout from `tsc` since the error messages are nasty, e.g.
  > tmp/ember_cli_typescript_fast_watcher-input_base_path- U6assc9z.tmp/web/tests/acceptance/search-test.ts(55,7): error TS2304: Cannot find name 'server'.


lib/typescript-fast-incremental-preprocessor.js
```diff
@@ -0,0 +1,94 @@
...
+        '--noEmit', 'false'
+      ], { silent: true }); // TODO: stdout is messy
+
+    // TODO: clean up process
+  }
+}
+
+function makeTempDir() {
+  // TODO: the directory won't get monitored if it's in tmp/
+  const dir = '.e-c-ts';
```
chriskrycho: This seems like a reasonable compromise to me; we'll just want to do two things if we run with it:

1. Document it.
2. Figure out how to automatically update `.gitignore` on installing the app, so that this directory is always excluded.

lib/typescript-fast-incremental-preprocessor.js
```diff
@@ -0,0 +1,94 @@
...
+    this.ui = options.ui;
+  }
+
+  toTree(inputNode, inputPath, outputPath) {
+    const js = funnel(inputNode, {
+      exclude: ['**/*.ts'],
+      annotation: 'JS files',
+    });
+
+    // TODO: how to disable watching .ts files?
```
chriskrycho: Once we figure out how to correctly identify the environment, it should either be available in the same way here *or* we can use it to set a flag on the `this` instance that we pass in so that we watch or not as appropriate for the environment.

Alternatively, maybe in the fast-compiler we intentionally just ignore these files entirely and only trigger on the files in the `.e-c-ts` directory changing?

dwickern:
I think the main problem are the two rebuilds
1. I modify `foo.ts`
1. ember-cli file watcher sees the change in `foo.ts` and triggers a rebuild
1. rebuild copies `foo.ts` to `tmp/.../foo.ts` using broccoli-funnel
1. `tsc --watch` file watcher sees the change in `tmp/.../foo.ts`
1. `tsc --watch` recompiles and outputs `.e-c-ts/foo.js`
1. ember-cli file watcher sees the change in `foo.js` and triggers a rebuild

It's a bit faster than broccoli-typescript-compiler but still not as fast as running `tsc --watch` to compile in-place. I think I need to take a different approach...

chriskrycho:
Ah, I misunderstood part of the flow in practice (a bit aspirationally I suppose). My assumption was (and I think if we could get there it would make sense if) it did:

1. Modify `foo.ts`.
2. `tsc --watch` recompiles and outputs `.e-c-ts/foo.js`.
3. ember-cli file watcher sees change in `foo.js` and triggers rebuild, which funnels to correct `tmp` directory.

The main challenge I'd foresee with that approach would be getting the bits from `.e-c-ts` into the correct `tmp` directory with the `funnel`, but I *think* you could map it correctly? (My broccoli knowledge is… not large.)

dwickern:
The advantage of using the preprocessor hooks is they have already sorted out and merged all the trees.

In your step 2, ember-cli also sees the change to `foo.ts` and triggers the second recompile. I think the only way to avoid this will be using a custom command, e.g. `ember serve-ts`

chriskrycho:
Mmmm. 🤔 @rwjblue, any suggestions here?

dwickern:
I changed this to use a separate command instead so that I can tweak the watcher

`ember serve-ts` 

It works! and it's _really fast_, as if you're not using typescript at all

chriskrycho:
Sweet! I’ll test it out and we can ship it!

chriskrycho:
These test failures are incredibly annoying. Nothing changed on our end for them. 😑

chriskrycho:
Two quick notes:

1. I pulled this down and tried to use it in our main app repository, and the TS watcher needs to exclude the blueprints directory. Otherwise, you end up with some nasty errors when using the substitutions that templates allow. We have a template that looks like this:

    ```
    import Component from '@ember/component';

    import { bem } from 'mobile-web/lib/utilities/bem';

    export const block = '<%= dasherizedModuleName %>';

    export default class <%= classifiedModuleName %> extends Component.extend({
      // `attributeBindings` and `classNameBindings` must be here until we have TS
      // support for the new decorator spec, as they need to be merged with other
      // items of the same name in the prototype chain.
      attributeBindings: [`elementId:data-test-${block}`],

      // You should remove `classNameBindings` and `tagClass` if you're going to
      // apply the BEM root to an element in the template explicitly.
      classNameBindings: ['tagClass'],
    }) {

      tagClass = bem({ block });
    }
    ```

    It's named `blueprints/bem-component/files/app/components/__name__.ts` (and there's a matching one for the integration test). Compilation falls over with this error:

    ```
    SyntaxError: mobile-web/blueprints/bem-component/files/app/components/__name__.js: Unexpected token (6:1)
      4 | export default class {
      5 | }
    > 6 |  %  > ;
        |  ^
      7 | Component.extend({
      8 |     // `attributeBindings` and `classNameBindings` must be here until we have TS
      9 |     // support for the new decorator spec, as they need to be merged with other
    ```

    It makes sense: it *is* a syntax error, but... we just need to exclude the `blueprints` directory.

2. We need to make sure we always delete the `.e-c-ts` directory after a run. Otherwise, problems like the one above will stick around after renames, etc.

chriskrycho:
Next comment: holy *smokes* this is fast.

chriskrycho:
Played with it a good bit; other than needing that fix for the blueprints directory and getting rid of the `.e-c-ts` directory, it's amazing.

So, to ready this for shipping it:

- [ ] add `.e-c-ts` to `.gitignore` on install
- [ ] clean up by deleting `.e-c-ts` when shutting down
- [ ] ignore the `blueprints` directory
- [ ] document the `serve-ts` command

dwickern:
Maybe `blueprints/` should be excluded in tsconfig? Or better yet, we could use explicit includes. I had similar issues before I disabled `allowJs` because it was compiling stuff in `server/`

dwickern:
We have actually 2 separate builds: running `tsc` directly (with `serve-ts`) and the broccoli pipeline build which does a bunch of transformations first. And both builds need to compile...

dwickern:
I also figured out why I couldn't put `.e-c-ts` into `tmp`: the directory is completely ignored in the watchman config.

chriskrycho:
@dwickern do you know what we need to do so we're not swallowing type errors when doing these incremental builds? I'd like to knock this out and ship it, and I *think* I know how to tackle the rest of it, but am unsure why those are getting swallowed at present. 🤔 Hopefully we can ship this as 1.1 this week, and ship the addon support soon as well.

dwickern:
I get an error on startup now
```
Error: process is not captured
    at Object.addHandler (/Users/dwickern/code/ember-cli-typescript/node_modules/ember-cli/lib/utilities/will-interrupt-process.js:82:13)
    at new Builder (/Users/dwickern/code/ember-cli-typescript/node_modules/ember-cli/lib/models/builder.js:37:30)
    at Class.run (/Users/dwickern/code/ember-cli-typescript/lib/serve-ts.js:41:21)
    at resolve (/Users/dwickern/code/ember-cli-typescript/node_modules/ember-cli/lib/models/command.js:317:20)
    at initializePromise (/Users/dwickern/code/ember-cli-typescript/node_modules/ember-cli/node_modules/rsvp/dist/rsvp.js:532:5)
```

chriskrycho:
Huh. I’ll straighten it out tomorrow. Not sure which of my changes would have caused that. 🤔

dwickern:
I bisected it to 1db94d52063e048d0af421f3fa2626fc33de7fcf

chriskrycho:
1. Thank you for hunting it down.
2. That’s… really weird. I’m not at my desktop at the moment and so can’t check myself easily; can you see if it breaks that way in 1.0.6 itself for you? (It did *not* for me, but that’s a bizarre change to have that effect!) If not, no worries; I’ll sort it out tomorrow.

dwickern:
The problem was resolving imports relative to e-c-ts instead of my project. For example, when I run `ember serve` it loads `my-project/node_modules/ember-cli/lib/models/builder.js`. When I run `serve-ts` it loads `ember-cli-typescript/node_modules/ember-cli/lib/models/builder.js`.

It should always load relative to my project.

dwickern:
Ctrl-C fails because `.e-c-ts` isn't empty:
```
Error: ENOTEMPTY: directory not empty, rmdir '.e-c-ts'
    at Object.fs.rmdirSync (fs.js:852:18)
    at Class.onInterrupt (/Users/dwickern/code/ember-cli-typescript/lib/serve-ts.js:87:10)
```

The message doesn't print the base url properly:
```
Build successful (145152ms) – Serving on http://localhost:4200undefined
```

chriskrycho:
Ahhhhh, good find.

Nice to know that `onInterrupt` *does* work here, though. Looks like we'll need to pull in `rimraf` or similar for cleanup?

dwickern:
I added back the tsc compiler output. The only problem is, `tsc --watch` detects changes inside `tmp/` which triggers several rebuilds and is quite noisy.

chriskrycho:
Hmm… maybe I'm failing to track correctly; shouldn't `tsc` *not* be looking at `tmp`?

chriskrycho:
A couple observations:

1. This way of doing the `node_modules` resolution definitely isn't working. 😒 (Not your fault; it's a weird thing.) I'm wondering if this is really just a problem with using `npm link`?
2. When using this in my local demo app (not my full size app), it's actually generating the `.js` files in place instead of generating them at the specified `--outDir` of `.e-c-ts`. That'll be the first thing I tackle.
3. I also need to add "ignore" rules to e.g. VS Code which starts trying to index `.e-c-ts` otherwise. 😱

blueprints/ember-cli-typescript/index.js
```diff
@@ -36,7 +37,23 @@ module.exports = {
...
+      // NOTE: this may or may not actually work -- it assumes that
+      // `"tmp:": true` is in fact set.
+      await this.insertIntoFile('.vscode/settings.json', '",\n.e-c-ts": true', {
+        after: '"tmp": true',
+      });
+    }
+
+    // TODO: same for Atom
+    // TODO: same for Sublime
+    // TODO: same for IntelliJ
```
dfreeman: Would this whole `afterInstall` hook be avoidable if the output directory lived in either system temp (`os.tmpdir()`) or piggybacked off the local broccoli temporary directory (which is presumably already set up to be ignored by the user's editor, git, npm, etc)?

chriskrycho: I would think so. And if we can piggy-back off of `os.tmpdir()`, that seems reasonable enough to me, though making sure *that* gets pulled into the Broccoli tree seems… iffy?

dwickern: We rely on watchman (or other watcher, but I've only tested with watchman) to pick up the compiled js.

Watchman is set up to monitor the root directory of the app, minus the excluded directories in the default `.watchmanconfig`:
```json
{
  "ignore_dirs": ["tmp", "dist"]
}

```

So we can't put the files in `tmp` or some other system directory

dwickern: It's possible we could add another watched directory somehow but I haven't explored the APIs for it

dfreeman: My understanding is it would end up adding that directory as a watchman root, but I'm pretty unfamiliar with the details there, so grain of salt. Best bet is probably just to try it and see what happens ¯\\\_(ツ)_/¯

.travis.yml
```diff
@@ -3,7 +3,7 @@ language: node_js
 node_js:
   - "6"
 
-sudo: false
+sudo: required
```
chriskrycho: Issue described [here](https://github.com/travis-ci/travis-ci/issues/9024) and [here](https://github.com/travis-ci/travis-ci/issues/8836#issuecomment-356362524) on Travis' end.

lib/serve-ts.js
```diff
@@ -0,0 +1,126 @@
...
+
+        this.ui.startProgress('Starting TypeScript compilation...');
+
+        // TODO: typescript might be installed globally?
+        // argument sequence here is meaningful; don't apply prettier.
+        // prettier-ignore
+        this.tsc = child_process.fork(
+          'node_modules/typescript/bin/tsc',
+          [
+            '--watch',
```
chriskrycho: @dwickern @dfreeman one thing I've been thinking is: we *should* be able to get and use the project's tsconfig, and then use that here (using `-p <path to project>`) except with overrides. I *believe* TS properly does the override sequence if you just hand it a project file and then a set of overrides.

Does that make sense to both of you?

chriskrycho: Alternative approach: do the usual `JSON.parse(tsConfig)` dance and set only the properties we want to set here, e.g. `sourceMap: <whatever is in the config` That might be better.

Also, I don't think we need to worry about using a global install of `tsc`; we should *only* be using the local version, since that's what will be used for the final build, as well.

dfreeman: I like the idea of using `-p <path-to-project>` and overriding with the handful of flags we need — it feels simpler (and I also _think_ `tsc` will actually let you get away with nonconformant JSON, so we'd need something fancier than `JSON.parse` for that approach)

dwickern: I think it already picks up the tsconfig in the project root based on the cwd? We could pass the path explicitly though.

chriskrycho: Yeah, I just went back and read the docs, and you're both correct about those pieces. So here, we should just be able to override *only what we need to* – and presumably, other than outdir, that should largely match what we have in the overrides in the `TypescriptPreprocessor` class (starting [here](https://github.com/typed-ember/ember-cli-typescript/pull/85/files#diff-fa6ce9add76b6ec544e7e53030f16ddcR24)).

Note that *that's* assuming valid JSON. 🤔

chriskrycho:
Two things:

1. Still seeing the problem of the tree getting pulled twice somehow. I end up seeing output like this for `ember serve`:

    ```
    WARNING: more-ts/Users/chris/dev/oss/more-ts/types/ember-cli-mirage/index.d.ts(220,16): error TS2451: Cannot redeclare block-scoped variable 'server'.
    types/ember-cli-mirage/index.d.ts(220,16): error TS2451: Cannot redeclare block-scoped variable 'server'.
   ```

    Best bet there is just that the `treeForApp` is getting something funky.

2. When I run `ember serve-ts` locally, I see: 

    ```
        /Users/chris/dev/oss/more-ts/node_modules/typescript/lib/tsc.js:58192
                    throw e;
                    ^

    TypeError: Cannot read property 'escapedText' of undefined
        at writeClassDeclaration (/Users/chris/dev/oss/more-ts/node_modules/typescript/lib/tsc.js:53824:84)
        at writeModuleElement (/Users/chris/dev/oss/more-ts/node_modules/typescript/lib/tsc.js:53428:28)
        at emitModuleElement (/Users/chris/dev/oss/more-ts/node_modules/typescript/lib/tsc.js:53389:17)
        at emitNode (/Users/chris/dev/oss/more-ts/node_modules/typescript/lib/tsc.js:54363:28)
        at emitLines (/Users/chris/dev/oss/more-ts/node_modules/typescript/lib/tsc.js:53136:17)
        at emitSourceFile (/Users/chris/dev/oss/more-ts/node_modules/typescript/lib/tsc.js:53335:13)
        at /Users/chris/dev/oss/more-ts/node_modules/typescript/lib/tsc.js:52928:17
        at Object.forEach (/Users/chris/dev/oss/more-ts/node_modules/typescript/lib/tsc.js:281:30)
        at emitDeclarations (/Users/chris/dev/oss/more-ts/node_modules/typescript/lib/tsc.js:52910:12)
        at Object.writeDeclarationFile (/Users/chris/dev/oss/more-ts/node_modules/typescript/lib/tsc.js:54418:37)
    ```

Just logging those here in case either of you gets to them ahead of me. I'll be at it again tomorrow.

PR #89, opened 2018-01-24
Solve 'The Mystery of the Missing app.js 🕵️‍♂️'

dfreeman:
Incremental compilation was broken because `app.js` was being transformed into a file just called `js`.

This should also fix things when using a linked copy e-c-ts. The `serve-ts` diff is nasty, but all I did was wrap the module in a function so it could use `project.require()` to ensure the CLI imports always resolve to the host app's copy.

### DIFF
--- a/index.js
+++ b/index.js
@@ -5,7 +5,7 @@ const fs = require('fs');
 const path = require('path');
 const SilentError = require('silent-error');
 const TsPreprocessor = require('./lib/typescript-preprocessor');
-const ServeTS = require('./lib/serve-ts');
+const buildServeCommand = require('./lib/serve-ts');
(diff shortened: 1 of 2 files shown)


chriskrycho:
OH MY GOSH. Hero.

dwickern:
Thanks for figuring this out! So we are importing relative to the project root instead of `process.cwd`.

(also FYI you can add `?w=1` to the github diff URL to ignore whitespace)

dfreeman:
> (also FYI you can add ?w=1 to the github diff URL to ignore whitespace)

This has to be about the fourth time I've learned that trick, but I always forget about it again by the next time I need it 🤦‍♂️

PR #93, opened 2018-02-07
Addon support

### LINKED ISSUE #47: Pre-publish step: build, generate typings
We need to make it possible for add-ons to use TypeScript without requiring consumers of add-ons to also install the 22MB dependency – esp. if add-ons are using multiple versions of TS, that will get out of hand quickly.

dfreeman:
Still need to work on the documentation, but installing off this branch and then running `ember g ember-cli-typescript` in a fresh addon should land you with a functional setup for publishing JS + declarations:
 - it generates a reasonable `tsconfig.json` for local development of the addon
 - it sets up `prepublishOnly` and `postpublish` scripts to deal with generating and then cleaning up the .js and .d.ts files

I successfully published versions of [`@dfreeman/why-yes-i-do-have-types`](https://github.com/dfreeman/why-yes-i-do-have-types) using both yarn and npm as I was playing with this, so at a bare minimum it works for one minimal test case 🙂 

Resolves #47.


chriskrycho:
Test failure is just AppVeyor being absurdly slow. Kicking it again.

chriskrycho:
Also: I AM SO EXCITED.

chriskrycho:
HEY IT WORKS. :shipit:

ro0gr:
That's awesome :heart:

PR #98, opened 2018-02-09
Add `helper` generator

tansongyang:


chriskrycho:
Thanks for this one as well, @tansongyang! One lint error here, too.

chriskrycho:
Follow-up for both this and the other: there’s something broken about this lint rule, I think. I’ll investigate further tomorrow, but it’s complaining about imports which are definitely used.

PR #99, opened 2018-02-09
Only run Appveyor against one Node/Ember version combo

dfreeman:
This simplifies our Appveyor setup not to run ember-try and to only test against one node version. It doesn't seem like ember-try is buying us a ton, since none of the build time components here are impacted by the Ember version, and I think if we want to get a more complete coverage matrix for Node versions, Travis is probably the simpler way to go.

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -4,17 +4,11 @@
 init:
   - git config --global core.autocrlf true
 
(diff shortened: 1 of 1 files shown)


chriskrycho:
I’ll merge it! …as soon as AppVeyor passes. :smirk:

PR #100, opened 2018-02-09
Add chalk, ember-router-generator as dependencies.

chriskrycho:
These are in the blueprints, so they need to be present when consumed. This wasn’t causing a problem before because, well… we didn’t have blueprints before!

This should resolve the lint problems in #96, #97, and #98.


chriskrycho:
Merging this and backporting to the others, even though AppVeyor hasn’t finished, b/c I suspect from what I was seeing that AppVeyor’s queue was *very* long tonight.

PR #102, opened 2018-02-09
More experimentation to shorten AppVeyor builds

dfreeman:
I'm going to use this PR as a way to keep experimenting with AppVeyor to see if we can find a way to shorten build times while still, y'know, working.

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -27,7 +27,7 @@ cache:
 test_script:
   # Output useful info for debugging.
   - yarn versions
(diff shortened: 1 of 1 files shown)


chriskrycho:
Ship it as soon as you're happy with it. I'm 100% in favor.

dfreeman:
Looks like this will run in 3-4 minutes instead of 15-20, so I'm calling it a win 🙂

chriskrycho:
:shipit: 👍

PR #104, opened 2018-02-10
Add `util` and `util-test`

tansongyang:
There are a couple of things different about this one.

1. I figured that it would reduce noise and clutter to start grouping similar generators together. So, I included both `util` and `util-test`. The generator for `util` was straightforward, just like the others. However ...
2. I got the error `Cannot find module '../test-framework-generator'` in `util-test-test.js`. I'm assuming this is because, in a real Ember app, the CLI is looking at what test framework you have installed and selecting the test file that way. I was hoping there would be another command line argument to specify test framework, but it doesn't look like there is. Any ideas on how to test? For the time being, I just left out tests for `util-test`.

### DIFF
--- a/blueprints/test-framework-detector.js
+++ b/blueprints/test-framework-detector.js
@@ -1,0 +1,47 @@
+'use strict';
+
+const fs = require('fs');
+const path = require('path');
+const VersionChecker = require('ember-cli-version-checker');
+
+module.exports = function(blueprint) {
+  blueprint.supportsAddon = function() {
+    return false;
+  };
+
+  blueprint.filesPath = function() {
+    let type;
+
+    let dependencies = this.project.dependencies();
+    if ('ember-qunit' in dependencies) {
+      type = 'qunit-rfc-232';
+
+    } else if ('ember-cli-qunit' in dependencies) {
+      let checker = new VersionChecker(this.project);
+      if (fs.existsSync(this.path + '/qunit-rfc-232-files') && checker.for('ember-cli-qunit', 'npm').gte('4.2.0')) {
+        type = 'qunit-rfc-232';
+      } else {
+        type = 'qunit';
+      }
+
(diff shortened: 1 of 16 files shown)


chriskrycho:
I *think* that’s actually a file we just need to bring over for the sake of making this work, but will confirm in the morning!

chriskrycho:
Never mind about the morning, apparently: [this](https://github.com/emberjs/ember.js/blob/master/blueprints/test-framework-detector.js) is the file we’ll need to bring over to make this work, @tansongyang. It should just go at the same place in *this* library as it does in Ember itself: the root of the blueprints directory.

tansongyang:
@chriskrycho Just glancing at the code, it looks like there's no way to test all styles (`qunit-rfc-232`, `qunit`, _and_ `mocha`). The code just checks your dependencies and chooses one for you, with no option to override. Are we OK with that?

chriskrycho:
We might actually just pull over the blueprint *tests* for these ones as well. (And maybe others! 🤔)

- [utils-test.js](https://github.com/emberjs/ember.js/blob/master/node-tests/blueprints/util-test.js)
- [utils-test-test.js](https://github.com/emberjs/ember.js/blob/master/node-tests/blueprints/util-test-test.js) (that name!)

tansongyang:
I think I got all the important files. `util-test` tests should work.

A couple of things I think are worth discussing:

1. Should I maybe make a different branch for adding the test framework code? This PR has a lot of stuff in it that might make more sense separated.
2. I wasn't aware that there were already blueprint tests in the `ember.js` repo. If I get to any other ones this weekend, I'll probably just be copying tests from there, especially for all the `*-test` generators.

chriskrycho:
1. Nah, I’m not really worried about it in this case.
2. Concur, this was going to be my broad suggestion. I didn’t think of it previously either.

PR #107, opened 2018-02-10
Add `helper` related generators

tansongyang:
This was my first go at a `*-addon` generator. Like `*-test`, it also required copying some files from the `ember.js` repo.

### DIFF
--- a/blueprints/helper-addon/files/__root__/__path__/__name__.ts
+++ b/blueprints/helper-addon/files/__root__/__path__/__name__.ts
@@ -1,0 +1,1 @@
+export { default, <%= camelizedModuleName %> } from '<%= modulePath %>';
--- a/blueprints/-addon-import.js
+++ b/blueprints/-addon-import.js
@@ -1,0 +1,54 @@
+'use strict';
+
+var stringUtil  = require('ember-cli-string-utils');
+var path        = require('path');
+var inflector   = require('inflection');
+
+module.exports = {
+  description: 'Generates an import wrapper.',
+
+  fileMapTokens: function() {
+    return {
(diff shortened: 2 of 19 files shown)


chriskrycho:
Thanks, @tansongyang! You're just knocking these out like crazy! To fix [the lint error](https://travis-ci.org/typed-ember/ember-cli-typescript/jobs/339953948#L495-L496):

```
/home/travis/build/typed-ember/ember-cli-typescript/blueprints/helper-test/index.js
  4:34  error  "ember-cli-is-package-missing" is extraneous  node/no-extraneous-require
```

Just add `"ember-cli-is-package-missing"` to the package.json. Easiest way I've been doing it is just `yarn add ember-cli-is-package-missing`.

chriskrycho:
Oh, and as a quick related note: you *should* be able to catch this kind of lint problem by doing `yarn run lint:js`. That'll drop your personal cycle time since we don't have to wait on Travis for it.

blueprints/helper-addon/files/__root__/__path__/__name__.ts
```diff
@@ -0,0 +1 @@
+export { default, <%= camelizedModuleName %> } from '<%= modulePath %>';
```
chriskrycho: @dfreeman is this kind of re-export going to work/compile properly as TypeScript? I *think* the answer we got to was *yes, everything works except things in `app/` in the root of an addon* but wanted to confirm.

dfreeman: I think we want these to just be `.js` files in the app tree — I can open a PR to update the blueprints we have in place

Issue #88, opened 2018-01-05
Add generator blueprint for prototype extensions


dwickern:
We should generate a `.d.ts` with prototype extensions.

Prototype extensions are controlled by `EXTEND_PROTOTYPES`:
https://guides.emberjs.com/v2.3.0/configuring-ember/disabling-prototype-extensions/

We can't conditionally enable types based on this so the best we can do is generate a `.d.ts` which the user can edit.

Array prototype extensions should be enabled by default for apps and disabled for addons. Function prototype extensions should be disabled by default probably. We haven't implemented string prototype extensions.

```ts
declare global {
    interface Array<T> extends Ember.ArrayPrototypeExtensions<T> {}
    interface Function extends Ember.FunctionPrototypeExtensions {}
}
```

dfreeman:
> Function prototype extensions should be disabled by default probably.

AFAIK array prototype extensions are the only ones that aren't already deprecated/[slated for deprecation](https://github.com/emberjs/rfcs/pull/272), so it makes sense to me that they'd be the only ones enabled by default.

dwickern:
Can a blueprint detect if it's an addon vs. app for the array prototype extensions?

dfreeman:
In most cases yes, though the addon story for e-c-ts is a little bit of an oddball, since having it as a `devDependency` means it's technically only installed for the addon's dummy app. We may have to settle for checking if the project name is `dummy` and basing the decision on that.

chriskrycho:
Looking at what I have for this in #108 and the discussion above: it looks like the function prototype extensions should probably be *included* but *commented out* to make it easy for people to “flip a switch” and turn them on?

HodofHod:
Just a heads up: enabling `interface Function extends Ember.FunctionPrototypeExtensions { }` causes typing errors when using `ember-decorators`' class decorators for me.

`[ts]
Unable to resolve signature of class decorator when called as an expression.
  Type 'Function' is not assignable to type 'void | typeof Badge'.
    Type 'Function' is not assignable to type 'typeof TooltipBadge'.`

Hope this helps anyone coming across that issue.

PR #108, opened 2018-02-11
Generate package typings

chriskrycho:
- Generates `types/<my-app>/index.d.ts` and puts the prototype extensions there if it's a normal app (#88)
- Moves types for `config/environment.js` to `types/<my-app>/config/environment.d.ts`
- Adds mapping to look up things in `types`

@dwickern @dfreeman all of this look sane?


blueprints/ember-cli-typescript/index.js
```diff
@@ -45,8 +54,24 @@ module.exports = {
...
+        return useAppDeclarations ? APP_DECLARATIONS : '';
+      },
+    };
+  },
+
+  fileMapTokens(/*options*/) {
+    // Return custom tokens to be replaced in your files.
+    return {
+      __app_name__(options) {
+        return options.dasherizedModuleName;
```
dfreeman: We want this to be `'dummy'` for addons, right? I think `dasherizedModuleName` will be the addon name here in that situation.

chriskrycho: This was the other bit I was wondering about, yeah.

chriskrycho: Updated to that effect, @dfreeman.

chriskrycho:
Concur: I think having them present and documenting it makes it *easy* for people to turn on if they need them, but makes the default do the "right thing" as it were. I'll merge as soon as AppVeyor goes green. Which will be a bit longer now that I've merged in `master` again, of course. 😭

PR #109, opened 2018-02-11
Use JS for addons' `app/` reexports

dfreeman:
I feel terrible undoing part of the hard work @tansongyang just did, but I _think_ for the `*-addon` blueprints, we actually don't want to override them at all. A simple `.js` reexport (the default implementation for those) should be exactly what we need.


chriskrycho:
Alas, I came to the same conclusion. And I blame *myself* for it; I should have checked every single generator and seen whether it needed it or not and documented it accordingly. 😖

tansongyang:
@dfreeman No worries! It happens. I'm just glad we got it sorted out.

PR #113, opened 2018-02-12
require dep on ember-cli-version-checker 2.1.0

csantero:
`blueprints/test-framework-detector.js` requires `ember-cli-version-checker`, but package.json doesn't specify a version. I don't think any ember app/addon is likely to be missing this dependency, since `ember-cli-babel` uses it, but this does mean that the version a developer has when they run the blueprint is likely to float.

This bit me because I encountered a bug when running e-c-ts generators in a private scoped addon. The bug is fixed in ember-cli-version-checker 2.1.0, so specifying the version in e-c-ts ensures that that bugfix is available for generators.

### DIFF
--- a/package.json
+++ b/package.json
@@ -44,6 +44,7 @@
     "ember-cli-string-utils": "^1.1.0",
     "ember-cli-test-info": "^1.0.0",
     "ember-cli-valid-component-name": "^1.0.0",
+    "ember-cli-version-checker": "^2.1.0",
     "ember-router-generator": "^1.2.3",
     "execa": "^0.9.0",
     "exists-sync": "^0.0.4",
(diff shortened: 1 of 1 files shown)


chriskrycho:
@csantero thank you! Since we don't use it directly, but simply include it via generated content, let's make it a peer dependency instead:

```json
{
  "peerDependencies": {
    "ember-cli-version-checker": "^2.1.0"
  }
}
```

That way we don't end up with problems from the curious ways yarn and npm try to deal with transitive dependency conflicts.

csantero:
@chriskrycho PR updated to use peer dep

chriskrycho:
Not sure why AppVeyor didn't get the message, but merging anyway.

PR #114, opened 2018-02-14
change e-c-version-checker from peer to regular dependency

csantero:


chriskrycho:
Thanks again, @csantero!

PR #123, opened 2018-02-16
Fix node-tests

dfreeman:
We need to be explicit about the ember-cli-qunit version we use in all acceptance tests — otherwise the test framework detector finds the version belonging to the project, which we shouldn't make assumptions about.

@chriskrycho 👋


chriskrycho:
Good catch. Also: yeeeesh.

PR #124, opened 2018-02-16
Support ember-cli-mirage out of the box

dfreeman:
This will cause Mirage support files to be included in the build using [the same logic](https://github.com/samselikoff/ember-cli-mirage/blob/master/index.js#L131) as ember-cli-mirage itself uses, as well as automatically setting up `paths` and `include` properly in the generator.


chriskrycho:
AppVeypr failure was a timeout. Kicked it to restart it.

chriskrycho:
Since CI passes, I’ll check it locally on my own app as well as then merge and cut a new release with it.

chriskrycho:
Confirmed it in my app!

PR #127, opened 2018-02-19
fix app path for addon in tsconfig

championswimmer:
for an addon, the `/app` folder is also present

the contents of the `/app` folder is merged into the namespace 
of the app that will consume this addon.

So the total namespace of the test app consists of both 
the `/app` folder as well as `/tests/dummy/app` folder

### DIFF
--- a/blueprints/ember-cli-typescript/index.js
+++ b/blueprints/ember-cli-typescript/index.js
@@ -51,7 +51,7 @@ module.exports = {
         }
 
         if (isAddon) {
-          paths[`${appName}/*`] = ['tests/dummy/app/*'];
+          paths[`${appName}/*`] = ['app/*', 'tests/dummy/app/*'];
         } else {
           paths[`${appName}/*`] = ['app/*'];
         }


dwickern:
Sounds good to me but I'll let Dan confirm

chriskrycho:
When we were first building this, we went back and forth here: add-on `app` directories are at this point *normally* only used for re-exports from the `addon` directory. What's the use case for other stuff? 🤔 (Not at all saying we *shouldn't* do this, just getting more info!)

championswimmer:
I cannot import from app namespace in my tests
like this 

`import isPromise from 'dummy/utils/is-promise';` 

I have to use 

`import isPromise from 'addon-name/utils/is-promise';`

In some cases we might want to use the former.

chriskrycho:
Hmm. As I understand it (and I might be misunderstanding something!), the point of `dummy` is precisely to test things as if the add-on is being used by an actual consuming app, in which case you'd never want to `import isPromise from 'dummy/...'` but instead `import isPromise from 'addon-name/...'`.

dwickern:
If the import resolves, then we should support it

championswimmer:
The import does resolve

championswimmer:
ALso, the default blueprint of `ember generate` creates unit tests that use 

`import x from 'dummy/y/x`

not 

`import x from 'addon/y/x`

so I believe this should be supported

chriskrycho:
I believe you *cannot* do that. You can supply multiple `tsconfig.json` files, but they're not *automatically* resolved the way ESLint configs are.

dwickern:
It would be nice if we could, since we have the same problem with e.g. app code trying to resolve test code.

championswimmer:
Closing in favour of https://github.com/typed-ember/ember-cli-typescript/pull/134

PR #134, opened 2018-02-20
Precompile TS files in addons' `app` trees

dfreeman:
In the absence of a way to compose a single TS project with multiple `tsconfig.json`s (see #127), this ensures we transpile any TS files found in addons' `app` trees when doing `ts:precompile`, and adds `app` to the `include` and `paths` setup for addons.

It _doesn't_ update all the blueprints to generate `.ts` files by default for the reexports 😖


dfreeman:
P.S. I don't know exactly what it looks like yet, but I'd love to see if we can figure out a way to get good test coverage in place for the `precompile` and `clean` commands.

championswimmer:
Given current situation, this sounds like a good solution to me

PR #135, opened 2018-02-20
Add TestContext type for this in test blueprints

jamescdavis:
This will generate tests such that tsc doesn't complain about `this` being untyped when `noImplicitThis` is turned on.


dwickern:
I'm not sure why the context isn't working for us given that the tests on DT pass:
https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/types/ember-qunit/ember-qunit-tests.ts

My hope is we can solve this without extra boilerplate

jamescdavis:
Oops, it IS working, I just didn't have `@types/ember-qunit`!

PR #137, opened 2018-02-20
Update addon generator to support sourcemaps.                                13:46:39

### LINKED ISSUE #129: how to get coverage reports
i was able to create an app and an addon using this,
my only concern is I am not able to get coverage reports to work with codecov/coveralls

Coverage for .ts files will need source maps. 

Any plans on how to support this ?

chriskrycho:
- Set the correct default in the `tsconfig.json` blueprint.
- Add a note in the README with details on how to enable support for sourcemaps in Ember CLI/Babel.

Resolves #129.

### DIFF
--- a/README.md
+++ b/README.md
@@ -7,6 +7,7 @@ Use TypeScript in your Ember 2.x and 3.x apps!
 * [Setup and Configuration](#setup-and-configuration)
   * [Ember Support](#ember-support)
   * [`tsconfig.json`](#tsconfigjson)
+  * [Sourcemaps](#sourcemaps)
 * [Using TypeScript with Ember effectively](#using-typescript-with-ember-effectively)
   * [Incremental adoption](#incremental-adoption)
   * [Install other types!](#install-other-types)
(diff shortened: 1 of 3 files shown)


README.md
```diff
@@ -84,6 +85,20 @@ However, there are a few things worth noting if you're already familiar with Typ
...
+### Sourcemaps
+
+To enable TypeScript sourcemaps, you'll need to add the corresponding configuration for Babel to your `ember-cli-build.js` file:
+
+```ts
+const app = new EmberApp(defaults, {
+  babel: {
+    sourceMaps: 'inline',
+  }
+}
```
dfreeman: Missing a paren 🙂

chriskrycho: Good catch!

dwickern:
The option is also documented at https://github.com/babel/ember-cli-babel#enabling-source-maps

I'm surprised source maps work at all without running into https://github.com/ef4/fast-sourcemap-concat/issues/39. Maybe the tsc output works around whatever bug causes babel to output invalid source maps?

Is there any performance impact enabling this? I'm not so concerned with typescript's performance but rather the rest of the broccoli pipeline.

chriskrycho:
@dwickern there's definitely a perf hit on the Babel side, though we were already paying that—and I added a note in the README to that effect. Running it locally, TS's generation of source maps seems to have little to no impact, which is nice.

PR #140, opened 2018-02-21
Test coverage for precompilation and tree merging

dfreeman:
This adds test coverage for the `ts:precompile` and `ts:clean` commands (and makes `precompile` clean up after itself).

It also introduces a pair of in-repo-addons to the dummy app and ensures that files in their `app` and `addon` trees will be built and are importable from the correct places.

### DIFF
--- a/package.json
+++ b/package.json
@@ -59,7 +59,7 @@
     "walk-sync": "^0.3.2"
   },
   "devDependencies": {
-    "@types/ember": "2.8.8",
+    "@types/ember": "2.8.13",
     "@types/ember-qunit": "^3.0.1",
     "@types/node": "*",
     "@types/qunit": "^2.0.31",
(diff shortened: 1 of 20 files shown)


chriskrycho:
It’s failing because of (…wait for it…) duplicate installations of the Ember typings. 🤦‍♂️

package.json
```diff
@@ -59,7 +60,9 @@
   },
   "devDependencies": {
     "@types/ember": "2.8.8",
+    "@types/ember-qunit": "^3.0.1",
```
chriskrycho: My guess is that this is referencing Ember’s types, and that’s how we’re ending up with the dupes. Probably need to throw a `resolutions` field in here. UGH.

chriskrycho: Confirmed, that’s the problem.

I *think* we’re okay if we add the resolutions field b/c it’s only for `devDependencies`… but I’m honestly not 100% confident in the semantics for yarn there. I’ll push a change with that and then we can ask for people to experiment with it in their own apps and addons.

chriskrycho:
@dfreeman I have bad news for you.

Windows and newlines are ANGRY ENEMIES.

One of us can fix that test tomorrow.

dfreeman:
🤦‍♂️ Should have anticipated that. I'll throw in some `trim()`s or `os.EOL`s tomorrow. Thanks for taking care of the type duplication!

PR #152, opened 2018-02-23
Ensure TS blueprints are used. Fixes #9.

### LINKED ISSUE #9: How to create controller in Typescript using class syntax?
Hi,

I'm trying this addon out in https://github.com/perlun/electron-ember-typescript-example/pull/3 but am running into some problems.

I am trying do define my controller like this:

```typescript
class ApplicationController extends Ember.Controller
```

...but the problem if I then export it using `export default ApplicationController`, it's actually a proper ES2015 _class_ that gets exported. It doesn't cause any _errors_, but the constructor never runs (presumably because it must be an `_init` method instead).

If I then export it as `export default Ember.Controller.extend(ApplicationController)`, it doesn't work either.

Has anyone done this successfully, and if so, do you have a valid example? (This should be added to the wiki or README file as well)

dfreeman:
Experimented with this locally. Making us be before `ember-source` and `ember-data` consistently resulted in JS files being generated, and putting us after gets us TS, so it looks like `after` is the ticket.

### DIFF
--- a/package.json
+++ b/package.json
@@ -105,6 +105,10 @@
     "before": [
       "ember-cli-babel"
     ],
+    "after": [
+      "ember-source",
+      "ember-data"
+    ],
     "paths": [
       "tests/dummy/lib/in-repo-a",
       "tests/dummy/lib/in-repo-b"


dfreeman:
Same! The only time I've ever run into it myself was when I constructed the situation by trying `before: ['ember-source']` for this PR.

PR #153, opened 2018-02-23
Update tsconfig when in-repo addons are generated

### LINKED ISSUE #146: Automatically update `tsconfig.json` when in-repo addons are generated
We automatically set up any in-repo addons we discover when the `ember-cli-typescript` generator runs, but any addons generated after that need some [manual config](https://github.com/typed-ember/ember-cli-typescript#in-repo-addons). It would be great to augment the `in-repo-addon` blueprint to do this without any user intervention.

jamescdavis:
resolve #146

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,10 @@ The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/) a
 
 ## [Unreleased]
 
+### Added
+
+* Blueprint (and tests) to generate in-repo addons configured for TypeScript
+
 ### Changed
 
 * Improve instructions for setting up [Linked Addons](README.md#linking-addons) and [In-repo Addons](README.md#in-repo-addons).
--- a/blueprints/ember-cli-typescript/index.js
+++ b/blueprints/ember-cli-typescript/index.js
@@ -2,6 +2,7 @@
 
 const fs = require('fs');
 const path = require('path');
+const updatePathsForAddon = require('../../lib/utilities/update-paths-for-addon');
 
 const APP_DECLARATIONS = `
 import Ember from 'ember';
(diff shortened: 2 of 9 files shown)


jamescdavis:
Yeah, I'm wasn't happy with` "ember-cli": "*"` in dependencies, either. I moved to `peerDependencies` as this also pleases eslint, yet shouldn't ever affect what version gets installed.

I would, though, rather have even this specify version, but didn't initially because I wasn't sure of the minimum version `ember-cli-typescript` supports? `"ember-cli": ">=2.12"`? But maybe should be `"ember-cli": "^2.12.0  || ~3.0.0"` and then bump versions as each new `ember-cli` minor release tests clean? Or could go with `"ember-cli": "^2.12.0  || ^3.0.0"` and just assume it'll work until 4.0 (because, ya know, we all follow semver perfectly, right? ;).

jamescdavis:
@dfreeman I'm always a fan of DRYing things ups, so I moved the code for updating `tsconfig.json` paths to a utility and used it both in the `in-repo-addon` and `ember-cli-typescript` blueprints. I also added an option to the `updatePathsForAddon` utility that will do the right thing for linked addons so that we can use this for a command that sets up linked addons.

I agree that an `ember-cli` peer dependency doesn't really help much, so I've switched to `eslint-disable-line node/no-unpublished-require`.

dfreeman:
@chriskrycho I know you're swamped this week, but if you happen to have a few minutes of downtime at some point, it looks like Appveyor just went out to lunch on the builds for this and #159 and never reported in, and I think you're the only person who can re-kick them.

chriskrycho:
Doh. Yeah, nice call out; I’ll kick them in a few.

chriskrycho:
Kicked. We'll see if it reports properly this time. If not, I'm just going to merge it, since it passes.

PR #159, opened 2018-02-26
Generate functioning addon components

jamescdavis:
Components generated in an addon are not currently functional because they are missing the manual setting of their layout property to their compiled template.

### DIFF
--- a/blueprints/component/index.js
+++ b/blueprints/component/index.js
@@ -63,7 +63,7 @@ module.exports = {
           'templates/components/' + stringUtil.dasherize(options.entity.name);
       }
       importTemplate   = 'import layout from \'' + templatePath + '\';\n';
-      contents         = '\n  layout';
+      contents         = '\n  layout = layout;';
     }
 
     return {
--- a/blueprints/component/index.js
+++ b/blueprints/component/index.js
@@ -62,7 +62,7 @@ module.exports = {
         templatePath   = pathUtil.getRelativeParentPath(options.entity.name) +
           'templates/components/' + stringUtil.dasherize(options.entity.name);
       }
-      importTemplate   = 'import layout from \'' + templatePath + '\';\n';
+      importTemplate   = '// @ts-ignore: Ignore import of compiled template\nimport layout from \'' + templatePath + '\';\n';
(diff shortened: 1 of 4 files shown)


blueprints/component/index.js
```diff
@@ -62,8 +62,8 @@ module.exports = {
         templatePath   = pathUtil.getRelativeParentPath(options.entity.name) +
           'templates/components/' + stringUtil.dasherize(options.entity.name);
       }
-      importTemplate   = 'import layout from \'' + templatePath + '\';\n';
-      contents         = '\n  layout';
+      importTemplate   = '// @ts-ignore: Ignore import of compiled template\nimport layout from \'' + templatePath + '\';\n';
+      contents         = '\n  layout = layout';
```
dfreeman: [nit] Can we add a semicolon after `= layout`, since the blueprints tend not to rely on ASI?

jamescdavis: oops!

jamescdavis:
Should I try to get [generation of a type declaration](https://embercommunity.slack.com/archives/C2F8Q3SK1/p1519666491000692) working? I'm not sure we have the types yet that we would need to do this. I'm pretty sure it should be [this](https://github.com/emberjs/ember.js/blob/master/packages/ember-glimmer/lib/template.ts#L19-L23), but I don't think that is published anywhere? Should I open a DefinitelyTyped PR to add it? Would it be appropriate to add a `@types/ember-glimmer` or just make it part of `@types/ember`?

chriskrycho:
Yeah, it's just not reporting. :sigh:

dfreeman:
@jamescdavis I don't think I can get back far enough in the Slack history for that link 🙂

As far as I know, the exact contents of compiled templates aren't public API, so if we did expose a type for them, I'd be in favor of just making it an interface with a brand field.

I do think generating declarations is an interesting idea. I'd love to have a chat at some point about how we can make generation work nicely throughout the ecosystem—things like ember-css-modules and ember-apollo-client (and I'm sure other addons) would likely benefit from having either something to plug into or an easy example to follow for producing declarations for their own generated modules.

PR #167, opened 2018-03-11
[WIP] convert to typescript

dwickern:
This is a first pass at converting the addon to typescript. I've converted the addon code, blueprints and supporting code. I haven't touched tests yet.

Most of the effort so far was spent writing [type definitions](https://github.com/typed-ember/ember-cli-typescript/tree/typescript/types) for various ember-cli/broccoli dependencies.


chriskrycho:
❤️ I'll try to take a gander if you want… but probably after EmberConf.

package.json
```diff
@@ -78,6 +90,7 @@
     "ember-cli-release": "^0.2.9",
     "ember-cli-shims": "^1.2.0",
     "ember-cli-sri": "^2.1.0",
+    "ember-cli-typescript": "^1.1.6",
```
championswimmer: I believe npm would error on this (same name as package) ?

chriskrycho: I haven't checked on publishing, but certainly none of the local operations complain. You'd never be able to publish _any_ package which wanted to use itself otherwise.

dwickern: Yes, for example typescript is written in typescript and thus has itself as a dependency. Inception!

championswimmer: @dwickern not really, they do not add it in package.json, I believe they use a last known good version here  https://github.com/Microsoft/TypeScript/tree/master/lib 

And yes that's what I meant @chriskrycho, it won't publish if same name is inside dependencies.

dwickern: It's there:
https://github.com/Microsoft/TypeScript/blob/master/package.json#L83

championswimmer: Okay this will not cause any problem. Releasing will work too as it is in devDependencies. I just checked with a dummy package :) 

Sorry for troubling :D

package.json
```diff
@@ -27,7 +27,9 @@
     "lint:js": "eslint ./*.js blueprints/*/index.js addon addon-test-support app config lib server test-support tests node-tests",
     "start": "ember serve",
     "test": "ember try:each",
-    "nodetest": "mocha node-tests --recursive"
+    "nodetest": "mocha node-tests --recursive",
+    "prepublishOnly": "ember ts:precompile",
```
dfreeman: Hadn't even occurred to me that this would work for our purposes here—nice! 😄

dfreeman:
What does development/testing of e-c-ts itself look like with this setup? Would we use `ts-node` or something like that?

dfreeman:
>  I think ultimately the build and the app will need separate tsconfigs.

Yep, that makes sense. It would be great once we've got this all figured out if we could make it as easy as possible for other addons to use TS in their node-side code.

Until MU lands, there would be some slight weirdness with declaration files (e.g. does `index.d.ts` refer to the build-time index or the runtime one?), but it still seems like a win overall—especially if we're able to get some of the library types you've written here upstreamed!

dwickern:
The root `index.d.ts` _should_ have typings for `index.js`, ie. build-time declarations.

chriskrycho:
@dwickern the problem is that it's overloaded: it also ends up being the root import for when you import something from an addon.

Maybe we should recommend doing `ember-cli-build.js` and `ember-cli-build.d.ts` for it?

dwickern:
Isn't that the build for the dummy app? 🙈

chriskrycho:
@dwickern thank you for starting this so long ago! ❤️ @dfreeman ended up carrying it forward during the push for 2.0, so I'm closing it now.

Issue #133, opened 2018-02-20
TS4082: Default export of the module has or is using private name 'Ember.Helper'


dfreeman:
When trying to generate declarations for an addon that exposes a helper with any of the 2.8 series of `@types/ember` releases, I see the following error:

```
TS4082: Default export of the module has or is using private name 'Ember.Helper'
```

To reproduce:
```sh
ember addon my-addon --yarn && cd my-addon
ember install ember-cli-typescript
ember generate helper foo
ember ts:precompile
```

chriskrycho:
How is it imported? Looks public [here](https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/types/ember/index.d.ts#L3433-L3437) and it's just a normal class [here](https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/types/ember/index.d.ts#L1340). 🤔

dfreeman:
```ts
import { helper } from '@ember/component/helper';

export function fooHelper(params/*, hash*/) {
  return params;
}

export default helper(fooHelper);
```

dfreeman:
There's been a [bunch](https://github.com/Microsoft/TypeScript/pull/22798) of [motion](https://github.com/Microsoft/TypeScript/pull/23351) in the past couple weeks to make the declaration emitter more helpful in these kinds of scenarios, but it looks like for names in _other_ modules, things are still open: https://github.com/Microsoft/TypeScript/issues/9944.

In the meantime, the workaround here is to change that import to `import Helper, { helper } from '@ember/component/helper';`, though that won't play nicely with the `noUnusedLocals` lint.

chriskrycho:
Oh, 🤦‍♂️ – I finally understand the privacy problem here. (For other readers who may have been confused: it's because in the specific context of the module `@ember/component/helper`, the class `Ember.Helper` itself isn't public – it's public elsewhere, and *another* type which is equivalent to it is also public, but that type is private in the module.

Ugh.

dfreeman:
This may or may not actually be fixed by more recent TS releases, but at some point we solved this particular problem by declaring the return type of `helper` in `@types/ember` to just be `any`.

[...]

PR #172, opened 2018-03-15
Invoke tsc programmatically

chriskrycho:


dzaman:
**TL;DR:** After installing `ember-cli-typescript` on an existing app I had trouble with automatic rebuilds / live reloading. Switching to the `programmatic-compilation` build solved this problem.

---

The initial build would work as expected in a normal amount of time for our app (20s), but rebuilds for me would take significantly longer (in some cases 5+ minutes) and in other cases I gave up waiting. I think it's possible that the completed builds were actually tied to additional file changes and would have otherwise hung indefinitely. 

When the rebuilds hung, I'd see output that a file has changed and that incremental compilation was starting, but neither the build success message nor the expected typescript warnings were output.

I would kill rebuilds older than that, but wasn't able to `ctrl+c` them, I'd have to `kill` the process. I'd see the "cleaning up..." message, but it would just hang after that. 

I'm using  `Ember 2.16.2` on `OS X Sierra 10.12.6`

jasonworden:
While trying @dzaman's branch with `ember-cli-typescript` on that same app, I had similar troubles. Here is what happened.
- Initial build took expected amount of time
- I was able to make 5-10 changes throughout a ~15 minute span that were reflected in the browser via hot-reloaded as expected.
- I made a subsequent change, and my browser tab for the app began to refresh but was hung up for several minutes. The re-build after these changes never finished.
- I ctrl-C'd out of the `ember server` process, but then had to `kill` the process.
- Chrome crashed, I believe right after I killed the server process.
- Restarted the server and was able to make a few changes with hot-reloaded working for about 5-10 minutes.
- A subsequent change did not result in a proper re-build, and I followed the same steps.
- I changed app branches and was able to work without the Typescript in the project.
- I'm working with the `programmatic-compilation` now and everything seems good!

`Ember 2.16.2` on `OS X Sierra 10.13.1`

jasonworden:
While trying @dzaman's branch with `ember-cli-typescript` on that same app, I had similar troubles. Here is what happened.
- Initial build took expected amount of time
- I was able to make 5-10 changes throughout a ~15 minute span that were reflected in the browser via hot-reloaded as expected.
- I made a subsequent change, and the browser tab did not reflect the changes. I manually refreshed the browser tab, and it was hung up trying to refresh for several minutes. The re-build after these changes never finished.
- I ctrl-C'd out of the `ember server` process, but then had to `kill` the process.
- Chrome crashed, I believe right after I killed the server process.
- Restarted the server and was able to make a few changes with hot-reloaded working for about 5-10 minutes.
- A subsequent change did not result in a proper re-build, and I followed the same steps.
- I changed app branches and was able to work without the Typescript in the project.
- I'm working with the `programmatic-compilation` now. Everything seemed good for awhile, but I eventually ran into the same issue as above.

`Ember 2.16.2` on `OS X Sierra 10.13.1`

chriskrycho:
@dfreeman one thing I was thinking about just now: can we leverage this to properly fail builds when people have the `noEmitOnError` setting enabled? That'd be a nice win here.

machty:
This PR fixes an [issue with ember-cli-addon-docs](https://github.com/ember-learn/ember-cli-addon-docs/issues/144) that I was running into where changing a `.md` file caused the rebuild to hang. I believe the root cause is that the `tsc` binary doesn't emit the "Compilation complete. Watching for file changes." text when it sees a non-JS/TS file change.

tansongyang:
Hello everyone. I'm checking in here to see what else needs to be done before this can be merged into `master`.

chriskrycho:
We need to get the tests passing, either by removing the helper-using test or by fixing the typings so it works properly (preferably but not *necessarily* the latter).

tansongyang:
@chriskrycho Is work already being done on the typings? If not, I think I could give some time to that this weekend.

chriskrycho:
Not to my knowledge. I haven’t had time to actually figure out what’s wrong there.

tansongyang:
OK. If I don't hear any updates by this weekend, I'll just start on it. Could you point me to where I should be looking?

chriskrycho:
@tansongyang what we know so far is documented in #133.

chriskrycho:
That is one super weird AppVeyor failure. Do we have some kind of caching problem, maybe?

dfreeman:
@chriskrycho Haha, I was just about to ping you to ask if we could try to clear that somehow. Bad cached dependency tree, maybe? 😬

jamescdavis:
:tada:

PR #173, opened 2018-03-16
add initial ember-cli-addon-tests

csantero:
The one test this adds:
* boots an app called `ts1` using ember-cli-typescript, with a pre-existing typescript file at `app/add.ts`
* downloads the final app javascript file
* ensures that the js file contains the expected transpiled output for a module called `ts1/add`


csantero:
Looks like the 5 minute timeout was too short for Appveyor. :(

node-tests/addon-tests/addon-test.js
```diff
@@ -0,0 +1,68 @@
...
+  after(function() {
+    return app.stopServer();
+  });
+
+  it('transpiles pre-existing typescript files', function() {
+    return request('http://localhost:49741/assets/ts1.js')
+      .then(response => {
+        expect(response.statusCode).to.equal(200);
+        let parsed = esprima.parseScript(response.body);
+        let bodyStatements = parsed.body
```
dfreeman: I wonder if we could get away with something like this to make it less scary to maintain/add more tests like this in the future?

```js
let parsed = esprima.parseScript(response.body);
let expected = esprima.parseScript(`
  export function add(a, b) {
    return a + b;
  }
`);

expect(parsed).to.deep.equal(expected);
```

csantero: @dfreeman Good idea! I'll add that in.

dfreeman:
@csantero Looks good! Re: `yarn` on Appveyor, I've had success by [adding an `install` entry for it](https://github.com/salsify/ember-css-modules/blob/master/appveyor.yml#L3) in the past (...apparently 😄)

chriskrycho:
Hey, sorry this sat for so long unattended, @csantero. AppVeyor *does* have yarn available in its test environment, so that shouldn't be the issue. Let's get this updated with latest from `master` and push it forward this week?

chriskrycho:
I'll see if I can poke at it next week if no one beats me to it.

Quick request for future reference: please add commits rather than rebasing for PRs to this repo. It's much easier to track what comments were in reference to that way! (No harm done, just figured I'd note it. Should put a note that way in a `CONTRIBUTING.md`, I think.)

csantero:
@chriskrycho interesting. I hate it when people update PRs against my repos with merge commits that bring in lots of commits from master. I guess it's personal preference?

chriskrycho:
Ha! Yes. I prefer that because it keeps the context. GitHub (and, for that matter, `git diff` and so on) is usually smart enough not to show changes pulled in that way.

csantero:
Also, my current theory about these failures is that they're just too slow. I don't think this PR is going to be viable until ember-cli-addon-tests sees some major performance improvements.

dfreeman:
We landed the moral equivalent of this work in #202, so I'm going to go ahead and close out this PR. Thanks for forging the path here, @csantero!

mike-north:
@dfreeman Are we sure that we have an addon acceptance test, not just an app one? I don't see it in the commits for #202

PR #180, opened 2018-04-07
Windows-compatible programmatic compilations

tansongyang:
So far, this contains just the changes needed to successfully compile on Windows. See commit comments for details on why I switched from `sane` to `chokidar`.

I tried to fix `ts:precompile`, but I wasn't entirely sure what it was supposed to do. Running `ember ts:precompile` on `master` right now produces this error:

```
Error: Command failed: tsc --outDir Me\AppData\Local\Temp/e-c-ts-precompile-23496 --rootDir Me\ember-cli-typescript --allowJs false --noEmit false --declaration --sourceMap false --inlineSourceMap false --inlineSources false

tests/dummy/app/helpers/typed-help.ts(7,1): error TS4082: Default export of the module has or is using private name 'Ember.Helper'.
```

So, I'll need some help fixing `ts:precompile`.

TLDR:

- [ ] Fix `ts:precompile`.
- [ ] Clean up. In particular, we should decide what we want `reportDiagnostic` and `reportWatchStatus` to do.

### DIFF
--- a/lib/utilities/compile.js
+++ b/lib/utilities/compile.js
@@ -3,7 +3,7 @@
 
 const mkdirp = require('mkdirp');
 const ts = require('typescript');
-const sane = require('sane');
+const chokidar = require('chokidar');
 const fs = require('fs-extra');
 
 module.exports = function compile(project, tsOptions, callbacks) {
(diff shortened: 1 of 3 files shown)


chriskrycho:
Heh, that specific failure looks like a problem with the type definitions, not with your work! (More on Monday.)

mike-north:
I ran into some mega annoying install issues due to chokidar's need for native dependencies, and 

This library was explicitly avoided in ember-cli, due to reasons stated in the following issue
https://github.com/ember-cli/ember-cli/issues/4503

Also react-native avoided adding it for similar reasons
https://github.com/facebook/react-native/issues/628

TL;DR - we may want to work our way back to using watchman

PR #194, opened 2018-04-19
Fix Windows path issue

tansongyang:
The fix turned out to be as simple as removing the offending line.

For posterity: originally, that line was added because the keys of `watchedFiles` all had forward slashes. That is no longer the case.


chriskrycho:
Per discussion on the other and the fact that this is actually *simpler*… I vote we :shipit:

tansongyang:
@chriskrycho @dfreeman I think we're all in agreement on merging this.

Issue #192, opened 2018-04-16
RFC: Type Generation


dfreeman:
I've been mulling over different approaches to this for a bit, and I'd love to start getting thoughts from the Typed Ember folks and the community in general 🙂 

## Background
There are a number of cases in a typical Ember app where files other than JS/TS sources produce importable modules in the compiled application. Because TypeScript is unaware of these files, it rejects attempts to import them. An elephant-in-the-room example of this is Handlebars templates, whose `import` statements [are `@ts-ignore`d in our component blueprint](https://github.com/typed-ember/ember-cli-typescript/blob/c7d4f923e3076bcc35e5fb511396a985dfb0cca4/blueprints/component/index.js#L65) to work around this exact issue.

The TypeScript compiler doesn't support any notion of loader plugins or anything like that to directly allow for deriving types from non-JS/TS files, but it does allow for [merging 'virtual directories' into a source tree](http://www.typescriptlang.org/docs/handbook/module-resolution.html#virtual-directories-with-rootdirs), which is the subject of this RFC.

## Proposal
I'd like to propose the following:
 1. we introduce a `types/generated` directory and add it to `rootDirs` and the appropriate places in `paths` to allow for both absolute and relative resolution of modules there
 2. we create a phase in the build process during which ember-cli-typescript can emit files into `types/generated` for use by `tsc` and other TypeScript-aware tooling
 3. we provide a mechanism for other addons to provide generated typings as well

For technical details on how these three points might work, see [Approaches](#Approaches) below.

### Templates

As an immediate win, I expect we would provide out-of-the-box behavior to produce type declarations for any templates found. The simplest approach here would likely be to produce a simple branded default export in order for component files to have _something_ to import, e.g.

```ts
// types/generated/addon/templates/my-component.d.ts
import { TemplateFactory } from 'htmlbars-inline-precompile';

declare const template: TemplateFactory;
export default template;
```

(Note that the `'htmlbars-inline-precompile'` module seems to be our source of truth for the type of a template at present...)

### `config/environment`

One non-obvious (or at least it wasn't immediately obvious to me) option we'd get from this is the ability to produce the type declaration for the `<app-name>/config/environment` module. The contents of this module are generated based on a node file executed at build time, and users are currently responsible for maintaining a one-off declaration file.

Instead, we may be able to inspect the shape of the config object during the build and emit a type declaration for it into the generated types directory.

### Addons

[...]

PR #196, opened 2018-04-23
Simplify watcher

dfreeman:
I worked some over the weekend on an implementation of #192, and in the process put the new watcher implementation through its paces. I found a couple scenarios (e.g. deleting a whole tree of directories) where the compiler would wind up hanging on to an orphaned reference to a file that no longer existed, and after comparing diagnostic output between our watcher and one of the default implementations, realized that their directory watcher was much more active than ours.

It seems with our chokidar setup we can forgo `watchFile` completely and just trigger the `watchDirectory` callback whenever a file within that directory changes. This eliminated all of the weird behavior I'd been seeing, and it got tsc's diagnostic output when using our watcher inline with what you'd see from the default one.

### DIFF
--- a/lib/utilities/compile.js
+++ b/lib/utilities/compile.js
@@ -1,11 +1,9 @@
 /* eslint-env node */
 'use strict';
 
-const ts = require('typescript');
 const chokidar = require('chokidar');
 const fs = require('fs-extra');
-const path = require('path');
-const debugWatcher = require('debug')('ember-cli-typescript:watcher');
+const debug = require('debug')('ember-cli-typescript:tsc:trace');
 
 module.exports = function compile(project, tsOptions, callbacks) {
   // Ensure the output directory is created even if no files are generated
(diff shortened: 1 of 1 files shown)


dfreeman:
@tansongyang given our history on with this particular code, if you happen to have a chance to try out this branch at some point before it's merged, that would be wonderful 🙂

tansongyang:
@dfreeman Apologies for the late response! I got back from our engineering offsite last night. I just tried this branch out and everything is A-OK.

chriskrycho:
:shipit:

lib/utilities/compile.js
```diff
@@ -14,9 +12,12 @@ module.exports = function compile(project, tsOptions, callbacks) {
   let fullOptions = Object.assign({
     rootDir: project.root,
     allowJs: false,
-    noEmit: false
+    noEmit: false,
+    diagnostics: debug.enabled,
+    extendedDiagnostics: debug.enabled
   }, tsOptions);
 
+  let ts = project.require('typescript');
```
jamescdavis: :+1:

dfreeman:
Great! I'll plan to merge and cut another beta release later today.

PR #202, opened 2018-04-29
Node acceptance tests, round 2

dfreeman:
Experimenting with a slightly different (hopefully faster?) approach to full-addon testing inspired by #173. Let's see how it plays in CI...


dfreeman:
Appveyor ran fine originally but seems to have gone out to lunch when I merged master?

PR #209, opened 2018-05-14
Stop inspecting output strings to determine when builds complete

dfreeman:
TypeScript 2.9 tweaked their diagnostic message when a build completes, which prompted me to finally go find a hook for reporting rebuilds that doesn't involve string sniffing. I tested this against 2.7, 2.8, and `next` and everything looked good across those three.

I'm also going to look at including different `typescript` versions in our build matrix, but figured I'd do that in a separate PR.

### DIFF
--- a/lib/incremental-typescript-compiler/index.js
+++ b/lib/incremental-typescript-compiler/index.js
@@ -97,14 +97,11 @@ module.exports = class IncrementalTypescriptCompiler {
 
     this._watchProgram = compile(project, { outDir, watch: true }, {
       watchedFileChanged: () => this.state.tscDidStart(),
+      buildComplete: () => this.state.tscDidEnd(),
 
       reportWatchStatus: (diagnostic) => {
         let text = diagnostic.messageText;
         debugTsc(text);
-
-        if (text.indexOf('Compilation complete') !== -1) {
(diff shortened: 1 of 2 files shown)


lib/incremental-typescript-compiler/index.js
```diff
@@ -97,14 +97,11 @@ module.exports = class IncrementalTypescriptCompiler {
 
     this._watchProgram = compile(project, { outDir, watch: true }, {
       watchedFileChanged: () => this.state.tscDidStart(),
+      buildComplete: () => this.state.tscDidEnd(),
```
chriskrycho: 😍

dfreeman:
I found a mention of `afterProgramCreate` in [the original PR for the watch API](https://github.com/Microsoft/TypeScript/pull/20234) and then confirmed its existence [here](https://github.com/Microsoft/TypeScript/blob/ba4bf21eade396ac0d82ab1278dfddc6ce69c653/lib/tsc.d.ts#L3363)

lib/incremental-typescript-compiler/index.js
jamescdavis: so much better!

PR #215, opened 2018-05-25
Remove `node_modules` from chokidar's ignore regex

dfreeman:
Excluding `node_modules` from watching was causing changes in linked addons not to be picked up. Removing that element from the exclude patterns fixes the problem, but I'm not positive why it was in there in the first place.

@tansongyang do you remember the context there? Were there Windows-related issues with allowing the watcher to include `node_modules`?

### DIFF
--- a/lib/utilities/compile.js
+++ b/lib/utilities/compile.js
@@ -67,7 +67,7 @@ function diagnosticCallback(callback) {
 function buildWatchHooks(project, sys, callbacks) {
   let root = escapeRegex(project.root);
   let sep = `[/\\\\]`;
-  let patterns = ['\\..*?', 'dist', 'node_modules', 'tmp'];
+  let patterns = ['\\..*?', 'dist', 'tmp'];
   let ignored = new RegExp(`^${root}${sep}(${patterns.join('|')})${sep}`);
 
   return Object.assign({}, sys, {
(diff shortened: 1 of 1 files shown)


tansongyang:
@dfreeman I don't recall touching this part of the file, so I can't say whether there are any Windows issues. I'm happy to test this PR on Windows, but I won't be available to do that until Sunday at the earliest.

dfreeman:
Looks like this is causing issues on Travis — `ENOSPC` in this context generally means the OS cap on watch handles has been reached. We could look at raising that in our Travis config, but it may indicate that we need to take a different tack rather than allowing _all_ of `node_modules` under the watcher.

dfreeman:
We likely do want to allow users to customize their ignored directories at some point (see #216 for an example use case), but for this particular issue I think the straightforward solution is to customize our `ignore` regex per watched root.

If we [ignore `node_modules` relative to the watched directory](https://github.com/typed-ember/ember-cli-typescript/pull/215/commits/bc198ef9d759d16031550350737e4986889df39d) rather than _always_ relative to the project root, then when tsc calls `watchDirectory('<project-root>/node_modules/linked-addon')`, we'll ignore `linked-addon`'s node_modules rather than that addon itself.

dfreeman:
Superseded by #222.

PR #240, opened 2018-07-22
Support for building test-support & addon-test-support

### LINKED ISSUE #175: Support `test-support` and `addon-test-support` code
Addons can expose code that's only available in test builds via `test-support` (which merges into `<app-name>/tests` and `addon-test-support` (which becomes `<addon-name>/test-support`). We should support transpiling from these two directories, as well as generating type declarations for publishing for the stuff in `addon-test-support`.

mike-north:
Fixes #175

### DIFF
--- a/index.js
+++ b/index.js
@@ -4,6 +4,7 @@
 const IncrementalTypescriptCompiler = require('./lib/incremental-typescript-compiler');
 const Funnel = require('broccoli-funnel');
 const MergeTrees = require('broccoli-merge-trees');
+const stew = require('broccoli-stew');
 
 module.exports = {
   name: 'ember-cli-typescript',
@@ -57,7 +58,8 @@ module.exports = {
       return this._super.treeForApp.call(this, tree);
     }
   },
-  // We manually invoke Babel for treeForAddon, treeForTestSupport and treeForAddonTestSupport
+
+  // We manually invoke Babel for treeForAddon and treeForAddonTestSupport
   // rather than calling _super because we're returning content on behalf of addons that aren't
   // ember-cli-typescript, and the _super impl would namespace all the files under our own name.
   treeForAddon() {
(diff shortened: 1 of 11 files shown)


mike-north:
[This addon](https://github.com/mike-north/ember-qunit-decorators) relies on the proposed change, in case anyone wants to test locally and poke around

mike-north:
There's still some stuff to do here. In particular
- [ ] `.d.ts` stuff has to land in the correct place (it's treated the same as the `/app` folder as things stand now)

mike-north:
ok, looks like this works with `ts:precompile` now :)

index.js
```diff
@@ -57,22 +57,33 @@ module.exports = {
       return this._super.treeForApp.call(this, tree);
     }
   },
-
+  // We manually invoke Babel for treeForAddon, treeForTestSupport and treeForAddonTestSupport
```
dfreeman: Is this true for `treeForTestSupport`? It looks like we're both invoking Babel _and_ calling `_super` there.

The default `treeForTestSupport` implementation in ember-cli looks like a no-op, so its Babel transpilation must be managed elsewhere (which is fine for us, since the `test-support` tree doesn't acquire the addon namespace), meaning we can probably skip Babel for that one.

dfreeman:
@mike-north I created a couple unfortunate merge conflicts for you in the main blueprint — I'm happy to sort through them if you'd like, or just let you handle it if you'd prefer.

index.js
mike-north: removed babel

mike-north:
> It would be great to get some basic test coverage ensuring that files in test-support and addon-test-support actually make it into the build under the expected module names.

There are no such tests for /addons, in-repo addons, etc... Refactoring the existing simple build test into something more comprehensive and flexible is beyond the scope of what I was aiming for with this PR. Are you open to "broader enhancement of acceptance tests" being an independent issue?

dfreeman:
I'm talking about [this file](https://github.com/typed-ember/ember-cli-typescript/blob/master/tests/unit/build-test.ts), which would cause the tests to fail to load if any of the expected modules were missing, and would fail individual assertions if the exported values weren't the expected ones.

mike-north:
@dfreeman unit test cases added. I thought you were referring to [the acceptance test](https://github.com/typed-ember/ember-cli-typescript/blob/master/node-tests/acceptance/build-test.js)

PR #323, opened 2018-10-01
fix: ember-cli-typescript-blueprints no longer claims highest priority

### LINKED ISSUE #322: [bug] - typescript blueprints do not respect addon ordering
```
ember-cli: 3.5.0-beta.1
node: 8.11.4
os: darwin x64
```

##### Setup
```sh
ember new sample; cd ./sample # Create a new ember app
ember install ember-cli-typescript # Set up typescript support
ember install ember-qunit-decorators
```
##### Repro
Run `ember g component-test x-foo`

Because of [the addon order](https://github.com/mike-north/ember-qunit-decorators/blob/2466c0d1fbf7d2b1afed5653c09290b4a99b14de/package.json#L82-L88), I'd expect to see the code generated by `ember-qunit-decorators`.

### What happened instead?
I see the code generated by `ember-cli-typescript-blueprints`

----

This problem is due to the fact that we do not respect addon ordering when inserting `ember-cli-typescript-blueprints` into the discovered addons array

https://github.com/typed-ember/ember-cli-typescript/blob/c0bc3d7460ea929ac18d0e0b2ee8b9218b358c82/ts/addon.js#L30-L32

The code as it currently stands, will result in the `e-cli-ts-blueprints` code generation winning out over all other addons.

We should probably find the index of `ember-cli-typescript` in the array, and splice the blueprints in immediately after it.

mike-north:
Fixes #322

### DIFF
--- a/package.json
+++ b/package.json
@@ -106,7 +106,13 @@
   "ember-addon": {
     "configPath": "tests/dummy/config",
     "before": [
-      "ember-cli-babel"
+      "ember-cli-babel",
+      "ember-qunit-decorators"
+    ],
+    "after": [
+      "ember-source",
+      "ember-data",
+      "ember-decorators"
     ]
   },
   "prettier": {
--- a/ts/addon.js
+++ b/ts/addon.js
@@ -29,7 +29,7 @@ module.exports = {
     // dependency on install.
     if (this.project.addons.includes(this)) {
       this.project.addons.splice(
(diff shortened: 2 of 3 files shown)


dfreeman:
Based on the test failures, I think we need to update [the `before` entry here](https://github.com/typed-ember/ember-cli-typescript/blob/c0bc3d7460ea929ac18d0e0b2ee8b9218b358c82/package.json#L108-L110) to have the same constraints as the blueprints.

PR #346, opened 2018-10-22
Refresh dependencies with ember-cli-update

dfreeman:
This updates CLI and friends to 3.5. Let's see if I got the Travis config right 🙂 

Edit: oh right, Github still isn't firing webhooks. Guess this will have to wait.


dfreeman:
I only seem to be able to get either Travis _or_ Appveyor to run against this branch, but they've each passed any time they did run 🤦‍♂️

PR #348, opened 2018-10-24
[v2] Code cleanup before an initial v2 prerelease

dfreeman:
This removes a bunch of v1 typechecking code and the corresponding dependencies, as well as updating our `engines` entry to match [the current recommendation for addons](https://github.com/ember-cli/ember-addon-output/blob/f9956d232011fc41e5ba128e978362ef9018b3e9/package.json#L54) and fixing a stray type declaration error.


chriskrycho:
@dfreeman is this meant to land before, after, or it-doesn't-matter with #347?

dfreeman:
These two can land in any order (but whichever is second will likely need to have `yarn.lock` merged)

PR #350, opened 2018-10-25
Maybe run Windows tests via Travis?

dfreeman:
Need to open a PR to see how this runs.

image


dfreeman:
Observations after playing with config and running things a few times:
 - the Windows containers still seem a bit slow/flaky, but definitely promising
 - testem doesn't seem to be able to find Chrome, and I'm not sure if that's because it's missing (the log says it's being installed) or if there's some other environmental thing going on

chriskrycho:
We can ask if anybody knows Travis folksiness could help diagnose (they use Ember IIRC so maybe we can just raise in #general in Discord?)

mike-north:
Shameless plug -- there's always Azure Pipelines (now free for OSS)
https://azure.microsoft.com/en-us/services/devops/pipelines/

mike-north:
Obsolete due to #353

PR #351, opened 2018-10-26
Ensure we don't break transpilation parallelizability

dfreeman:
From https://github.com/emberjs/data/pull/5647/files#r228593401

Prior to this we were configuring the TypeScript Babel plugin as `[pluginPath]`, but `broccoli-babel-transpiler` assumes if you specify an array that you intend to include a second configuration parameter, and `undefined` isn't a JSON-serializable value.


mike-north:
confirmed to work w/ ember-data

PR #359, opened 2018-11-01
Update README for v2 `ember install`

mike-north:
v2 won't function properly unless set up as a `dependency`. ember-cli installs packages as `devDependency` by default (the correct default imo).

This updates the readme to clarify how the code on master should be installed into apps.

### DIFF
--- a/README.md
+++ b/README.md
@@ -81,7 +81,7 @@ For issues relating to typescript compiler analysis, [create an issue in this pr
 To install or upgrade the addon, just run:
 
 ```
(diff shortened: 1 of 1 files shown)


dfreeman:
The code on `master` is still v1 (v2 lives on [the `v2` branch](https://github.com/typed-ember/ember-cli-typescript/tree/v2)) — also, `--save` is only needed when installing as an addon dependency; we might want to clarify that to avoid confusion.

PR #385, opened 2018-11-12
Semantic Release & Conventional Commits

mike-north:
This PR adds validation of [conventional commits](https://conventionalcommits.org) and versioning/release automation via [semantic-release](https://semantic-release.gitbook.io/) to ember-cli-typescript.

We have a sharable configuration here

https://github.com/typed-ember/semantic-release-config

that includes

- automatic changelog generation
- automatic version incrementing, based on conventional commits
- automatic github release publish, tagging & notifying issues upon relevant commits being added to a release
- automated npm publish

---

One thing to consider

- We can have the script automatically commit the changelog and updated version number in package.json to master. semantic-release is very good about doing this _only_ if no new commits are on master since the commit being tested. **this would involve removing branch protection on the master branch**

- If we want to leave branch protection in place, the changelog will still be in the "releases" page, since that involves only pushing a tag and creating a release (neither of which involves touching `master`). If we want to go this way, the convention would be to set the `package.json` version to `0.0.0-development`.


.travis.yml
```diff
@@ -1,54 +1,66 @@
...
   chrome: stable
-
 cache: yarn
-
 env:
   global:
-    # See https://git.io/vdao3 for details.
     - JOBS=1
+    # npm (ZYtF)
+    - secure: ZYtFfhPoZX+TVkGicbLaWmeL1btgHaY8BZj+IuEMFJJuCft0Q0ixOxFZ/e/9GjLfEZp0wjNSjE1TgQmJagH4V0Tis7cGD2Gv2V4ZHMgvR7UI9Ktwa+s4HYZs2Q6rhTHsXWG9EJ6cFW74x3Kjly8a49RoG4tXGeaZBjq3ANgTuLF6gBHTSTtSzwFYcyj2R7csF+1k7xFNf/mFUh+YTtpnUshng54UiYxk/Yz+tOQAYLF0tYlSiMpbwwHc2uwxaoZcJ2J1lkyHsP8nCYM18cLX6CGaueqbAVDNlBEDljgwb4Hnt0omiBOeR9iS/elNEDwAddERav8sIdR0IdGJxKEeg9HzEfSXd8Y0Wh/XB/u86c2WgTKyHW6hidUigVZD3uyudSjBc+ZBwc7zb7dhMWfEozYs1yC9JkiYKyI9TdIu24imqZwE1eCm3+vVVn0Qo/zcFFueUOZNp/cL884hJ4iRV4ed65PacVWzsf1/Q5ThiYYi2XRgNW9q1RR1V51BGQx2zEeBnnP4rwGzfExXHyLmsN76uTTUMONRoxAPIMfrAK0EQfiCTi6hfk7JgqUsgS3sdNbLTDcvqwtq0K09CRCnT51iUqZqlohH8QVMtv9SaWIKVpPHV1BICfxMdXIpdc4dmNKnEmYXNtUTfWRNb4EzVCA2j77T6emDrNZBwgHbHJY=
```
mike-north: dedicated npm account for publishing

.travis.yml
```diff
@@ -1,54 +1,66 @@
...
+    - stage: 'Tests @ Locked Dependencies'
+      name: 'Tests @ Locked Dependencies'
+      install:
+        - yarn install --non-interactive # omitted usual --no-lockfile
       script:
-        - test $MD_ONLY && echo "Skipped!" || yarn lint:js
-        - test $MD_ONLY && echo "Skipped!" || yarn ci:prepare
-        - test $MD_ONLY && echo "Skipped!" || yarn ci:test
-    - stage: "Versioned Tests"
+        - test $MD_ONLY && echo "Skipped! 'yarn lint:js'"    || yarn lint:js
```
mike-north: `lint:js` only needs to be run once total. I've chosen the earliest job possible (either in the first test stage would be fine) to minimize pointless builds and maximize time-to-feedback

.travis.yml
```diff
@@ -1,54 +1,66 @@
...
     - env: EMBER_TRY_SCENARIO=ember-cli-beta
-
   include:
-    - stage: "Tests"
-      name: "Tests"
+    # Scenario: obey lockfile
+    - stage: 'Tests @ Locked Dependencies'
+      name: 'Tests @ Locked Dependencies'
+      install:
+        - yarn install --non-interactive # omitted usual --no-lockfile
```
mike-north: this represents "locked" dependencies. If this test ever passes and "Floating dependencies" fails, it means that a newly released dependency version may break users.

.travis.yml
```diff
@@ -1,54 +1,66 @@
...
-        - test $MD_ONLY && echo "Skipped!" || yarn lint:js
-        - test $MD_ONLY && echo "Skipped!" || yarn ci:prepare
-        - test $MD_ONLY && echo "Skipped!" || yarn ci:test
-    - stage: "Versioned Tests"
+        - test $MD_ONLY && echo "Skipped! 'yarn lint:js'"    || yarn lint:js
+        - test $MD_ONLY && echo "Skipped! 'yarn ci:prepare'" || yarn ci:prepare
+        - test $MD_ONLY && echo "Skipped! 'yarn ci:test'"    || yarn ci:test
+
+    # Scenario: lockfile deleted & re-created
+    - name: 'Floating Dependencies'
```
mike-north: this represents the transitive dependencies that a consumer would get _right now_, since library lockfiles mean nothing to consumers.

.travis.yml
```diff
@@ -1,54 +1,66 @@
...
+        - test $MD_ONLY && echo "Skipped! 'yarn ci:test'"    || yarn ci:test
+
+    - stage: 'Supported TypeScript & Ember-CLI Versions'
       env: EMBER_TRY_SCENARIO=typescript-release
     - env: EMBER_TRY_SCENARIO=typescript-beta
     - env: EMBER_TRY_SCENARIO=ember-cli-release
+      node_js: 10 # slightly more modern node, so travis-deploy-once selects this job for publish
+      script:
+        - test $MD_ONLY && echo "Skipped!" || yarn ci:prepare
+        - test $MD_ONLY && echo "Skipped!" || yarn ember try:one $EMBER_TRY_SCENARIO && yarn travis-deploy-once "yarn semantic-release"
```
mike-north: Because this test is run against a slightly more modern node version, it will always be selected by https://github.com/semantic-release/travis-deploy-once as the one that does the publish. Because of this, it's also the only one that bothers to cleanup after `ember-try` stuff is done.

.travis.yml
```diff
@@ -1,54 +1,66 @@
...
+        - test $MD_ONLY && echo "Skipped! 'yarn ci:test'"    || yarn ci:test
+      env: EMBER_TRY_SCENARIO=typescript-beta
+    - env: EMBER_TRY_SCENARIO=ember-cli-beta
 before_install:
   - bash ./is_md_only.sh && MD_ONLY=true && echo "Only .md files have changed!" || test true
-
 install:
-  - test $MD_ONLY && echo "Skipped!" || yarn install
-
+  - test $MD_ONLY && echo "Skipped!" || yarn install --no-lockfile --non-interactive
```
mike-north: by default, we shouldn't be using this library's lockfile to obtain dependencies, since our consumers won't care about the `e-cli-ts` lockfile.

.travis.yml
```diff
@@ -1,54 +1,66 @@
...
+        - test $MD_ONLY && echo "Skipped!" || yarn ci:prepare
+        - test $MD_ONLY && echo "Skipped!" || yarn ember try:one $EMBER_TRY_SCENARIO && yarn travis-deploy-once "yarn semantic-release"
     - env: EMBER_TRY_SCENARIO=ember-cli-beta
 
+    # Scenario: Things we keep an eye on, but do not technically support.
+    #           This stuff should align with jobs->allow_failures above
+    - stage: 'Canary Tests'
+      script:
+        - test $MD_ONLY && echo "Skipped! 'yarn ci:prepare'" || yarn ci:prepare
+        - test $MD_ONLY && echo "Skipped! 'yarn ci:test'"    || yarn ci:test
```
mike-north: Same test script that `'Supported TypeScript & Ember-CLI Versions'` runs, but no semantic-release. This way, `travis-deploy-once` only runs on jobs that _must_ pass in order for a release to go out.

.travis.yml
```diff
@@ -1,54 +1,66 @@
...
+  - test $MD_ONLY && echo "Skipped!" || yarn install --no-lockfile --non-interactive
 script:
-  # Typecheck and compile the addon, then nuke the TS source so we're testing the code we would publish
   - test $MD_ONLY && echo "Skipped!" || yarn ci:prepare
-  # Usually, it's ok to finish the test scenario without reverting
-  #  to the addon's original dependency state, skipping "cleanup".
-  - test $MD_ONLY && echo "Skipped!" || yarn ember try:one $EMBER_TRY_SCENARIO --skip-cleanup
-
-# We build PRs, but don't trigger separate builds for the PR from the branch.
+  - test $MD_ONLY && echo "Skipped!" || yarn ember try:one $EMBER_TRY_SCENARIO --skip-cleanup && yarn travis-deploy-once "yarn semantic-release"
```
mike-north: This script will only run for `'Supported TypeScript & Ember-CLI Versions'` scenarios _other than the one that does the actual library publish_, because all other jobs have a `script:` explicitly specified.

package.json
```diff
@@ -123,5 +130,26 @@
...
+    "tag": "latest"
+  },
+  "husky": {
+    "hooks": {
+      "commit-msg": "./node_modules/.bin/commitlint -e $HUSKY_GIT_PARAMS"
+    }
+  },
+  "toolchain": {
+    "node": "8.12.0",
+    "yarn": "1.10.0"
```
mike-north: https://github.com/notion-cli/notion

.travis.yml
jamescdavis: Good catch!

.travis.yml
```diff
@@ -1,54 +1,66 @@
----
 language: node_js
 node_js:
-  - "6"
-
+  - 8
```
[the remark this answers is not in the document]
dfreeman: I'm not particularly worried about testing with node 8 at this point—as you said, it's only got about six months left, and we have TS to transpile recent syntax like `async`/`await`. As long as we avoid any APIs that came about in 7 or 8 (and I'm not aware of anything that's particularly tempting in that band), we should be fine.

(Also worth noting that our `chokidar` dependency is gone in v2)

mike-north:
Closing this, as our CI setup as changed considerably over the past six months

PR #388, opened 2018-11-13
[v2] Smarter declaration placement for `ts:precompile`

dfreeman:
In switching to Babel for transpilation in `v2` we've eliminated a lot of the implicit assumptions about Ember project structure in ember-cli-typescript's internals, so now any TypeScript code that makes it through a project's normal preprocessor pipeline is guaranteed to be transpiled.

The `ts:precompile` command, though, has still required hard-coded knowledge of the typical addon layout up to this point. This is problematic for packages like [ember-test-helpers](https://github.com/emberjs/ember-test-helpers/pull/450) that don't adhere to a normal project structure, requiring a [postprocessing step](https://github.com/emberjs/ember-test-helpers/blob/27389be9e42b6f1c2c295f8c8d3ba6517f1a80f6/scripts/move-declaration-files.js) after running `ts:precompile` to move the declarations into the correct location.

This PR removes all knowledge of that hard-coded structure from the precompile command, instead reading the `paths` hash in `tsconfig.json` and moving any declarations whose mapping begins with the project's package name into the corresponding location on disk. This means `ember-test-helpers` will be supported out of the box in 2.0, and it also makes it so that module unification is no longer a special case—in both situations the correct behavior falls out naturally from the project's `paths` config.


ts/lib/commands/precompile.ts
```diff
@@ -0,0 +1,72 @@
...
+
+  _loadConfig(outDir: string) {
+    let ts = this.project.require('typescript') as typeof import('typescript');
+    let configPath = ts.findConfigFile(this.project.root, ts.sys.fileExists);
+    if (!configPath) {
+      throw new Error('Unable to locate `tsconfig.json`');
+    }
+
+    let configSource = ts.readJsonConfigFile(configPath, ts.sys.readFile);
+    let config = ts.parseJsonSourceFileConfigFileContent(configSource, ts.sys, path.dirname(configPath));
```
mike-north: It's a little difficult to identify the correct way to read config files. 

https://github.com/mike-north/code-to-json/blob/a1395ce8c478dfc00856e10cbba11f711e057640/packages/cli/src/commands/run.ts#L76-L78 worked for me. Some subset of the approaches I tried failed to handle `{ extends: "foo" }` properly, so that's probably worth testing

ts/lib/commands/precompile.ts
```diff
@@ -0,0 +1,72 @@
...
+      '--declarationDir', outDir,
+      '--rootDir', rootDir || this.project.root,
+      '--allowJs', 'false',
+      '--noEmit', 'false',
+      '--declaration',
+      '--emitDeclarationOnly',
+      '--sourceMap', 'false',
+      '--inlineSourceMap', 'false',
+      '--inlineSources', 'false',
+    ]);
```
mike-north: I'm not sure what the reasoning is behind hardcoding these options, rather than using what's found the appropriate tsconfig (if one is found), overriding `--emitDeclarationOnly`. 

**A use case**  - if I have strictness settings defined, and `noEmitOnError` true, I'd expect no declaration files to be generated unless the strict type-checking works.

We also should consider `--declarationMap` (TS 2.9+)

ts/lib/commands/precompile.ts
dfreeman: This properly follows `extends` chains, and I generally prefer it over `readConfigFile` as it has better type guarantees (you get a `ParsedCommandLine` rather than `any`).

ts/lib/commands/precompile.ts
dfreeman: This was carried over from [the previous implementation](https://github.com/typed-ember/ember-cli-typescript/pull/388/files#diff-f11fe69778211644d8597f7e5f089f58L27) (the file has just changed enough over the commits here that it's treated as a delete/create in the full diff).

The flags here are layered on _over_ what's in `tsconfig.json`, the same as they would be when invoking `tsc` by hand, so the scenario you're describing already works exactly as you'd expect. In fact, several of these flags are only necessary in case their opposite is set in `tsconfig.json` (e.g. `tsc` complains if you have both `--declaration` and `--allowJs`, so we have to explicitly turn the latter off).

dfreeman: I do agree that investigating the new declaration sourcemaps stuff is a good idea, though it may be that we can just allow users to configure that in their `tsconfig.json` and not need to set it explicitly. Now that we're not dealing with `.ts` -> `.js` transpilation here, we can actually likely drop the sourcemap related flags entirely and let the user do what they like. I'll update that file accordingly.

dfreeman:
Thanks for dealing with that merge conflict @mike-north 😄

PR #464, opened 2018-12-07
[v2] Check off a couple remaining todo items

dfreeman:
- Include `isolatedModules: true` in the default `tsconfig.json` as [recommended by Microsoft when transpiling with Babel](https://blogs.msdn.microsoft.com/typescript/2018/08/27/typescript-and-babel-7/)
- Warn when we detect ember-cli < 3.5, because project-local `tmp` can upset `tsc`

### DIFF
--- a/blueprint-files/ember-cli-typescript/tsconfig.json
+++ b/blueprint-files/ember-cli-typescript/tsconfig.json
@@ -3,6 +3,7 @@
     "target": "es2017",
     "allowJs": true,
     "moduleResolution": "node",
+    "isolatedModules": true,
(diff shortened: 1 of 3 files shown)


mike-north:
I've seen quite a few issues come in relating to `isolatedModules: true`. Seems like a rough DX edge for some users -- we can smooth this out via "above the fold" info in release notes.

TL;DR
> It's important to enable `isolatedModules` when using babel for TS compilation, because it catches, and provides early feedback for scenarios that `tsc` is fine with, but will make babel fail.

PR #493, opened 2018-12-24
Add some information about uninitialized class properties

mfeckie:
Got burned by unintialized class properties, which relates to https://github.com/babel/babel/pull/9141

### DIFF
--- a/README.md
+++ b/README.md
@@ -414,6 +414,20 @@ export default class UserProfile extends Component {
 
 This also holds true for all other macros of the ember-decorators addon.
 
+#### Typing uninitialized properties
+
+For folks using babel@7 you may encounter problems when providing type definitions such as the following when working with Ember below `3.6`.
+
+```ts
+import Component from '@ember/component';
+
+export default class UserProfile extends Component {
+  username?: string;
+}
+```
+
+When invoked via a template `{{user-profile username='example123'}}`, you would expect that `username` would have the value of `example123`, however prior to the native class feature released in Ember `3.6`, this will result in `username` being undefined.  
+
 #### Ember Data lookups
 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -414,6 +414,22 @@ export default class UserProfile extends Component {
...
+
+```ts
+import Component from '@ember/component';
+
+export default class UserProfile extends Component {
+  username?: string;
+}
+```
+
+When invoked via a template `{{user-profile username='example123'}}`, you would expect that `username` would have the value of `example123`, however prior to the native class feature released in Ember `3.6`, this will result in `username` being undefined.
```
mike-north: We should consider whether it's preferable to just include `ember-native-class-polyfill` as either a dependency or something we check for in this addon's `included()` hook. Looks like the polyfill supports _at least_ Ember 3.4 (https://github.com/pzuraq/ember-native-class-polyfill/issues/11) which satisfies our framework version support policy.

The root cause of the problem this documentation is trying to help users avoid, is that babel 7 auto-initializes uninitialized properties to `void 0` (possibly for perf reasons, so runtimes see instances as having a consistent shape), which interferes with the way we expect `Ember.Component` (and `Glimmer.Component`) to work.

igbopie:
The polyfill might be the solution for me as I ran into issues with ember upgrade.

README.md
buschtoens: Initializing the field as `void 0` / `undefined` has the side-effect of being good for perf, but Babel primarily does this to be spec-compliant.

https://github.com/babel/babel/issues/9105 and  https://github.com/babel/babel/pull/9141 go into detail about this.

chriskrycho:
@mfeckie I'm going to rebase this on Monday so I can incorporate these notes into my docs PR (#599) and get it landed as I update the docs. Thank you again!

chriskrycho:
GitHub is confused; this landed via 28059eb9. Thanks again @mfeckie!

PR #593, opened 2019-02-26
fix: warn when addons under development have .ts files in app/

### LINKED ISSUE #562: Cannot override a service exported with typescript file
ember-cli: 3.7.1
node: 8.12.0
os: linux x64

### Please paste the output of `tsc -v` here
tsc -v
The program 'tsc' is currently not installed. You can install it by typing:
sudo apt install node-typescript

### Please paste your `tconfig.json` and `tslint.json` or `eslint.json` (if applicable) below
<details><summary><b>My tsconfig.json</b></summary><pre>

```
{
  "compilerOptions": {
    "target": "es2017",
    "allowJs": true,
    "moduleResolution": "node",
    "allowSyntheticDefaultImports": true,
    "noImplicitAny": true,
    "noImplicitThis": true,
    "alwaysStrict": true,
    "strictNullChecks": true,
    "strictPropertyInitialization": true,
    "noFallthroughCasesInSwitch": true,
    "noUnusedLocals": true,
    "noUnusedParameters": true,
    "noImplicitReturns": true,
    "noEmitOnError": false,
    "noEmit": true,
    "inlineSourceMap": true,
    "inlineSources": true,
    "baseUrl": ".",
    "module": "es6",
    "paths": {
      "repro-ajax-ts/tests/*": [
        "tests/*"
      ],
      "repro-ajax-ts/*": [
        "app/*"
      ],
      "*": [
        "types/*"
      ]
    }
  },
  "include": [
    "app/**/*",
    "tests/**/*",
    "types/**/*"
  ]
}
```
  
</pre></details>

1. ember new repro-ajax-ts
2. ember install ember-cli-typescript@next
3. Bump ember-ajax to 4.0.2
4. Create a service named ajax.js that extends ajax service from ember-ajax

##### Reproduction Case

Reproduction: https://github.com/Gaurav0/repro-ajax-ts (Run ember test)

Expected test to pass; i.e. was able to override ajax service from addon

### What happened instead?

Test fails; ajax service is the one from the addon; not the app.

TLDR: if an addon exports a service using a .ts file in its app directory, an app using it cannot override it using a .js file

Related issue: https://github.com/ember-cli/ember-ajax/issues/428

dfreeman:
Closes #562. This doesn't change the underlying behavior, but it should give addon authors a heads up in cases where they're likely to run into it. Notably, the warning isn't emitted for in-repo addons, since their host typically has ember-cli-typescript set up as well, so their app trees will be processed.

### DIFF
--- a/ts/addon.ts
+++ b/ts/addon.ts
@@ -8,6 +8,8 @@ import fork from './lib/utilities/fork';
 import TypecheckWorker from './lib/typechecking/worker';
 import TypecheckMiddleware from './lib/typechecking/middleware';
 import { Application } from 'express';
+import walkSync from 'walk-sync';
+import fs from 'fs-extra';
 
 export default addon({
   name: 'ember-cli-typescript',
@@ -16,6 +18,7 @@ export default addon({
     this._super.included.apply(this, arguments);
     this._checkDevelopment();
     this._checkPeerVersions();
+    this._checkAddonAppFiles();
 
     // If we're a direct dependency of the host app, go ahead and start up the
     // typecheck worker so we don't wait until the end of the build to check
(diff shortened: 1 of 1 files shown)


ts/addon.ts
```diff
@@ -17,8 +17,8 @@ export default addon({
   included() {
     this._super.included.apply(this, arguments);
     this._checkDevelopment();
-    this._checkPeerVersions();
-    this._checkAddonAppFiles();
+    this._checkAddonAppFiles();y
```
chriskrycho: You've an extra `y` here, @dfreeman!

chriskrycho: …looks like you fixed it. 😆

ts/addon.ts
```diff
@@ -124,6 +127,27 @@ export default addon({
...
+    // ...and are at the root of the project (i.e. not in-repo)...
+    let isRootAddon = this.parent.root === this.project.root;
+
+    // ...and have .ts files in their `app` directory.
+    let appDir = `${this.parent.root}/app`;
+    if (isDevelopingAddon && isRootAddon && fs.existsSync(appDir)) {
+      let tsFilesInApp = walkSync(appDir, { globs: ['**/*.ts'] });
+      if (tsFilesInApp.length) {
+        this.ui.writeWarnLine(
+          `found .ts files in ${appDir}\n` +
```
chriskrycho: Can we elaborate this and/or tweak the wording a little? I'd like for us to make clear that the `.ts` file *won't be compiled*, so e.g. imports will fail.

ts/addon.ts
dfreeman: Yep, that was the result of resolving a merge conflict on my phone 🙃

ts/addon.ts
dfreeman: You mean like on the next line where it says we _only_ compile files in `addon/`? 😜

But in seriousness, do you have a suggestion for phrasing here that will make it clearer?

chriskrycho: Yeah, I kept going back and forth on the wording and whether the next line was clear enough and I kept landing on “somebody is going to fail to understand this.” Still thinking on it; haven’t gotten anywhere. Let’s ship it as is and if one of us comes up with a better phrasing we can update it.

PR #597, opened 2019-03-06
Remove resolutions rule for '@types/handlebars'.

chriskrycho:
Since the DefinitelyTyped issue which triggered the need for this hack has now been resolved, we can remove this hack. This will unblock other dependency upgrades, since the lockfile will once again work nicely.

### DIFF
--- a/package.json
+++ b/package.json
@@ -111,7 +111,8 @@
     "typescript-eslint-parser": "21.0.2"
   },
   "resolutions": {
-    "@types/ember": "3.0.28"
+    "@types/ember": "3.0.28",
+    "@types/ember__string": "3.0.6"
   },
   "engines": {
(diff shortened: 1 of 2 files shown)


package.json
```diff
@@ -111,8 +111,7 @@
     "typescript-eslint-parser": "21.0.2"
   },
   "resolutions": {
-    "@types/ember": "3.0.27",
-    "@types/handlebars": "npm:handlebars@^4.1.0"
+    "@types/ember": "3.0.27"
```
jamescdavis: Don't you mean?
```suggestion
    "@types/ember": "3.0.28"
```

chriskrycho: Why yes, yes I do! 😆 😭 Fixed.

PR #643, opened 2019-03-29
Types for compiled template modules

### LINKED ISSUE #242: [discussion] How should we handle templates?
Currently, importing a template (i.e., in an addon's component) includes a 

##### lib/my-addon/addon/components/x-foo.ts
```ts
// @ts-ignore: Ignore import of compiled template
import layout from '../templates/components/x-foo';
```
to get around the issue of typescript being unable to resolve the module. Since we're already maintaining a lot of bookkeeping for addons (in particular, `tsconfig.json` paths/includes for in-repo addons/engines), have we considered 

##### lib/my-addon/addon/components/x-foo.ts
```ts
import layout from 'my-addon/templates/components/x-foo';
```
along with
##### (A) types/my-addon.d.ts
```ts
declare module 'my-addon/templates/*' {
  import { TemplateFactory } from "htmlbars-inline-precompile";
  export default TemplateFactory
}
```
or 

##### (B) types/my-addon.d.ts
```ts
declare module 'my-addon/templates/*' {
  export default any
}
```

or 

##### (C) types/my-addon.d.ts
```ts
declare module 'my-addon/templates/*' {
  export default any
}
```

mike-north:
Deliberately based off of #644, which is based off of #647
Fixes #242 
Related #192

This PR results in a `types/global.d.ts` file being generated, producing a module type augmentation for paths we expect to contain templates

```ts
  import { TemplateFactory } from 'htmlbars-inline-precompile';
  const tmpl: TemplateFactory;
  export default tmpl;
```

Depending on project layout the path of the module will be different

*Classic*
```ts
declare module 'foo/templates/*' { ... }
```
*Pods*
```ts
declare module 'foo/components/*/template' { ... }
```
*Module Unification*
```ts
declare module 'foo/ui/components/*/template' { ... }
```

### DIFF
--- a/ts/blueprints/ember-cli-typescript/index.js
+++ b/ts/blueprints/ember-cli-typescript/index.js
@@ -3,16 +3,36 @@
 const fs = require('fs');
 const path = require('path');
 
-const APP_DECLARATIONS = `
-import Ember from 'ember';
+const APP_DECLARATIONS = `import Ember from 'ember';
 
 declare global {
   interface Array<T> extends Ember.ArrayPrototypeExtensions<T> {}
   // interface Function extends Ember.FunctionPrototypeExtensions {}
 }
 
-export {};
+export {};`;
+
+/**
+ * @param {string} projectName
+ * @param {'classic' | 'pods' | 'mu'} layout
+ */
+function buildTemplateDeclarations(projectName, layout) {
+  const comment = '// Types for compiled templates';
+  const moduleBody = `
+  import { TemplateFactory } from 'htmlbars-inline-precompile';
+  const tmpl: TemplateFactory;
+  export default tmpl;
 `;
+  switch (layout) {
+    case 'classic': return `${comment}
+declare module '${projectName}/templates/*' { ${moduleBody}}`;
+    case 'pods': return `${comment}
+declare module '${projectName}/components/*/template' { ${moduleBody}}`;
+    case 'mu': return `${comment}
+declare module '${projectName}/ui/components/*/template' { ${moduleBody}}`;
+    default: throw new Error(`Unexpecte project layout type: "${layout}"`);
+  }
+}
 
 module.exports = {
   APP_DECLARATIONS,
@@ -33,6 +53,7 @@ module.exports = {
     let isAddon = this.project.isEmberCLIAddon();
     let isMU = this._detectMU();
     let includes = isMU ? ['src'] : ['app', isAddon && 'addon'].filter(Boolean);
+    const isPods = this.pod;
 
     includes = includes.concat(['tests', 'types']).concat(inRepoAddons);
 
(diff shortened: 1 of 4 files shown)


ts/blueprints/ember-cli-typescript/index.js
```diff
@@ -3,16 +3,36 @@
 const fs = require('fs');
 const path = require('path');
 
-const APP_DECLARATIONS = `
-import Ember from 'ember';
+const APP_DECLARATIONS = `import Ember from 'ember';
```
mike-north: unrelated fix to avoid an unnecessary newline before the `types/my-app/index.d.ts` imports

ts/blueprints/ember-cli-typescript/index.js
```diff
@@ -3,16 +3,36 @@
...
+  const moduleBody = `
+  import { TemplateFactory } from 'htmlbars-inline-precompile';
+  const tmpl: TemplateFactory;
+  export default tmpl;
 `;
+  switch (layout) {
+    case 'classic': return `${comment}
+declare module '${projectName}/templates/*' { ${moduleBody}}`;
+    case 'pods': return `${comment}
+declare module '${projectName}/components/*/template' { ${moduleBody}}`;
```
dfreeman: I think this needs to account for `podModulePrefix`—it will also miss any route templates that are defined. Given how loose things are with pods, the simplest approach might just be `declare module '${projectName}/*/template' {`

ts/blueprints/ember-cli-typescript/index.js
```diff
@@ -3,16 +3,36 @@
...
+    case 'mu': return `${comment}
+declare module '${projectName}/ui/components/*/template' { ${moduleBody}}`;
+    default: throw new Error(`Unexpecte project layout type: "${layout}"`);
```
dfreeman: Typo

```suggestion
    default: throw new Error(`Unexpected project layout type: "${layout}"`);
```

ts/blueprints/ember-cli-typescript/index.js
```diff
@@ -33,6 +53,7 @@ module.exports = {
     let isAddon = this.project.isEmberCLIAddon();
     let isMU = this._detectMU();
     let includes = isMU ? ['src'] : ['app', isAddon && 'addon'].filter(Boolean);
+    const isPods = this.pod;
```
dfreeman: Since the default blueprint is normally run as part of `ember install`, I believe this flag will only ever be set if we have users explicitly re-run the generator afterwards with `ember g ember-cli-typescript --pod`. Is that something we want to encourage?

mike-north: I tested this manually with the `.embercli.js` file having the pods flag set to true. The scenario I'm thinking of is


```shell
ember new i-should-not-use-pods --pods # sets up .embercli.js with "pods": true
cd i-should-not-use-pods
ember install ember-cli-typescript

```

ts/blueprints/ember-cli-typescript/index.js
mike-north: Thanks! I'll go with your "simplest approach" idea

ts/blueprints/ember-cli-typescript/index.js
dfreeman: Gotcha; I'd forgotten you could set a default there

PR #644, opened 2019-03-30
Convert node tests to TS

mike-north:
Possibly through a subtle change in imports, this fixed some non-determinism in tests as well (i.e., we have Renovate PRs that are failing for strange and seemingly unrelated reasons).


ts/types/ember-cli/index.d.ts
```diff
@@ -88,5 +95,11 @@ declare module 'ember-cli/lib/models/project' {
...
     isEmberCLIAddon(): boolean;
     require(module: string): unknown;
+    isModuleUnification(): boolean;
   }
 }
+
+declare module 'ember-cli-blueprint-test-helpers/helpers';
+declare module 'ember-cli-blueprint-test-helpers/chai';
+declare module 'ember-cli-blueprint-test-helpers/lib/helpers/ember';
+declare module 'mktemp';
```
dfreeman: Can we move these declarations out of the ember-cli types? Might make them easier to find if we want to come flesh them out in the future.

PR #647, opened 2019-03-30
Azure pipelines - linux & compatibility testing

mike-north:
- Browser tests in firefox & chrome, across windows and linux
- TS & Ember-CLI compatibility test scenarios in Azure


chriskrycho:
@mike-north have you been able to figure out why our Windows Azure builds keep timing out?

mike-north:
@chriskrycho I'll need to get the azure folks involved to figure out what's going on. Sometimes the tests pass (i.e., the current azure linux build for this PR), sometimes the ember tests can't connect to headless chrome within the 30-second window, and sometimes our "skeleton app" can't be deleted in our mocha tests.

ts/tests/helpers/skeleton-app.ts
```diff
@@ -0,0 +1,133 @@
...
+  teardown() {
+    if (this.watched) {
+      this.watched.kill();
+    }
+    this.tmpDir.removeCallback();
+  }
+
+  _ember(args: string[]) {
+    let ember = require.resolve('ember-cli/bin/ember');
+    return execa('node', [ember].concat(args), { cwd: this.root, cleanup: true });
```
mike-north: Potentially important change, `, cleanup: true` option ensures that killing the process also kills any child processes that it has spawned.

dfreeman: I think this option already [defaults to `true`](https://github.com/sindresorhus/execa#cleanup)

mike-north:
> Is the plan to move entirely to Azure and drop Travis/Appveyor once this lands?

I know that ember-cli (and possibly ember) are investigating the feasibility of doing this. I think it's still a good idea to run them side-by-side for a while until we figure out what's best for this project

ts/tests/helpers/skeleton-app.ts
mike-north: good catch!

mike-north:
Merging based on @dfreeman's approval

PR #649, opened 2019-04-01
Drop support for node 6 (EOL)

mike-north:
Deliberately based on a bunch of other PRs

* #647 
   * #644 
      * #643
         * this PR

Node.js 6.x reaches its end of life on `2019-04-30`

The main things we get by merging this are:
 * Native async/await support in build/server code
 * We no longer need `co`, since it was being used exclusively for async chains in tests
 * Compiled TS output (blueprints, tests) can be ES2017 instead of ES2016

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -70,7 +70,6 @@ module.exports = {
       },
       rules: {
         'node/no-unpublished-require': 'off',
-        'node/no-unsupported-features/es-syntax': 'off',
       },
     },
 
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -15,5 +15,6 @@
     "ember-beta": true,
     "ember-canary": true,
     "ember-default": true
-  }
+  },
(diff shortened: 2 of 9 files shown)


dfreeman:
The addon ecosystem generally treats dropping support for a version of Node (even an EOL one) as a breaking change — would we release this as ember-cli-typescript@3? Are there any other changes we'd like to land along with this one?

mike-north:
> The addon ecosystem generally treats dropping support for a version of Node (even an EOL one) as a breaking change

Agree

> would we release this as ember-cli-typescript@3? Are there any other changes we'd like to land along with this one?

Can't think of any

chriskrycho:
Seems sensible to me as well. I don't see anything else *breaking* that's meaningful to land between now and the end of the month; the other things on our roadmap are all additive.

mike-north:
This can go in now that we're past the Node6 EOL date

PR #651, opened 2019-04-01
Refactor azure config

mike-north:
Rather than having three config files, this allows us to have one. 

We also now get a separate CI status for each job in the build matrix (i.e., each TS version, each ember-cli version).

For efficiency reasons, the fixed-dependency linux build runs first. If it succeeds, the remaining jobs will run in parallel with a maximum concurrency of 5

Linting is only done on the linux builds (both fixed and floating dependencies), since we have no reason to expect different results in other scenarios


package.json
```diff
@@ -34,7 +34,7 @@
     "ci:test": "yarn ci:log-version-info && yarn ci:test:app && yarn ci:test:node",
     "ci:test:app": "ember test",
     "ci:test:node": "mocha --recursive js/tests",
-    "prepublishOnly": "yarn tsc --project ts --noEmit false",
+    "prepublishOnly": "yarn tsc --noEmit false --project ts",
```
chriskrycho: Any particular reason to flip this invocation order?

PR #667, opened 2019-04-09
Update toolchain to latest Node LTS, Yarn.

chriskrycho:
Just what it says on the tin. If you're using [notion](https://www.notionjs.com), this will make sure your tools are in sync with the project's.

### DIFF
--- a/package.json
+++ b/package.json
@@ -146,6 +146,7 @@
     "proseWrap": "never"
   },
   "toolchain": {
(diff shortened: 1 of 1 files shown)


chriskrycho:
This is only failing on beta. I'm going to merge, but open an issue tracking our failure against beta. cc. @mike-north @dfreeman @jamescdavis

PR #678, opened 2019-04-13
chore(tools): add prettier with prettier-eslint

chriskrycho:
This PR enables Prettier in our codebase, with the same rules we had in place before… but in such a way that they'll actually be automatically enabled and enforced.

In support of that change:

- I switched us to using the newly official `@typescript-eslint` packages.
- I integrated `eslint-plugin-prettier`, which both checks Prettier configuration as part of our lint phase and provides users with an easy workflow (`yarn lint --fix`) to resolve it from the command line, as well as solid editor integration across all editors.
- I actuallly *ran* Prettier across the whole codebase with the same rule set as before.
- I noticed that we still have the old `.jshintrc` file as I was tweaking other lint configuration files, and so removed it.
- I saw that we were getting *weird* ESLint errors out of our generated files during test runs, and remembered that we're already linting independently, and that Ember itself is likely to drop `ember-cli-eslint` from the default blueprint soon, so I removed `ember-cli-eslint` as well.


.eslintrc.js
```diff
@@ -1,16 +1,18 @@
 module.exports = {
   root: true,
-  parser: 'babel-eslint',
+  parser: '@typescript-eslint/parser',
   parserOptions: {
     ecmaVersion: 2017,
     sourceType: 'module',
   },
-  plugins: ['ember'],
+  plugins: ['ember', '@typescript-eslint', 'prettier'],
```
dfreeman: Does including the plugin without enabling any of the rules do anything?

package.json
```diff
@@ -76,12 +76,13 @@
     "@types/resolve": "0.0.8",
     "@types/semver": "6.0.0",
     "@types/tmp": "0.1.0",
+    "@typescript-eslint/eslint-plugin": "^1.6.0",
     "babel-eslint": "10.0.1",
```
dfreeman: If we're using `@typescript-eslint/parser` everywhere now, can we drop `babel-eslint` from our devDeps?

ts/types/ember-cli-blueprint-test-helpers/index.d.ts
```diff
@@ -1,6 +1,8 @@
 declare namespace Chai {
   interface ChaiStatic {
-    file(name: string): {
+    file(
+      name: string
+    ): {
```
dfreeman: A little surprised by this one 🤔

ts/tests/blueprints/ember-cli-typescript-test.ts
```diff
@@ -136,9 +138,7 @@ describe('Acceptance: ember-cli-typescript generator', function() {
 
         const globalTypes = file('types/global.d.ts');
         expect(globalTypes).to.exist;
-        expect(globalTypes)
-          .to.include("declare module 'my-addon/templates/*'")
-          .to.include(`
+        expect(globalTypes).to.include("declare module 'my-addon/templates/*'").to.include(`
```
jamescdavis: IMHO, this is harder to read than what we had. What rule is forcing this?

ts/tests/blueprints/ember-cli-typescript-test.ts
```diff
@@ -206,9 +206,8 @@ describe('Acceptance: ember-cli-typescript generator', function() {
 
           const globalTypes = file('types/global.d.ts');
           expect(globalTypes).to.exist;
-          expect(globalTypes)
-            .to.include("declare module 'my-app/ui/components/*/template'")
-            .to.include(`  import { TemplateFactory } from 'htmlbars-inline-precompile';
+          expect(globalTypes).to.include("declare module 'my-app/ui/components/*/template'").to
+            .include(`  import { TemplateFactory } from 'htmlbars-inline-precompile';
```
jamescdavis: This one's even worse. Can we tweak the rules to allow this?

ts/tests/unit/copy-declarations-test.ts
```diff
@@ -66,74 +62,72 @@ describe('Unit: copyDeclarations', function() {
...
+        },
+      },
     });
 
-    expect(createdNodes).to.deep.equal([
-      'addon-file.d.ts',
-      'file.d.ts',
-      'generated-file.d.ts'
-    ]);
+    expect(createdNodes).to.deep.equal(['addon-file.d.ts', 'file.d.ts', 'generated-file.d.ts']);
```
jamescdavis: Why? IMHO, this is not better.

ts/types/ember-cli-blueprint-test-helpers/index.d.ts
jamescdavis: Yeah, that's odd.

package.json
chriskrycho: Oh, good call – I think so, yes. I'll do that and confirm everything still works.

ts/types/ember-cli-blueprint-test-helpers/index.d.ts
chriskrycho: I'm surprised by this one also. Heh.

ts/tests/blueprints/ember-cli-typescript-test.ts
chriskrycho: This is just Prettier's default behavior with maximizing use of line length. This kind of thing is why Prettier strongly recommends the maximum length of 80; I had set it to 100 to match other projects I was working on back when I included a config a long time ago, but I have no objection to doing it at 80 instead. (I actually considered doing that here but opted to leave the settings as they were and just actually enable it.)

ts/tests/blueprints/ember-cli-typescript-test.ts
chriskrycho: I'll take a look. We definitely can't be the first ones to run into `.to.<some method>` as a thing to preserve with it.

ts/tests/unit/copy-declarations-test.ts
chriskrycho: Again, just line length.

chriskrycho:
I'll note to the general sentiments that I also *often* disagree with specifics around Prettier's choices; but I've found the value of just not having to worry about it is pretty much always more than worth it. I'll poke and see if I can tweak a couple specific things – I'll also push a comparison (which we can roll back if we choose) with 80-character width limits.

ts/tests/blueprints/ember-cli-typescript-test.ts
jamescdavis: Looks like it's the multiline template literal that's causing this. If it were not for that, prettier will do [this](https://prettier.io/playground/#N4Igxg9gdgLgprEAuEcAeAHOYYAoDmANhAEYCGhAKgJ5YDOAlAHQwRMCWUYhArgCZxcAHRADuZAE5wABAFsIfHoRkByWdQC0ZPn2gB6eLIyEy8OnoBUKkc1YcuvAbgAGAKx50Y06DMKc4zgwA3CAANCAQGDDs0HTIoJISEADuAAqSCHEoZABuEOx8YSAkEmRgANZwMADKGGWc+MgwEjxw4QAWMLKEAOrt7GZ1YHDVmQPsOQPUyOB0ceGcdHASMKml+LJkyABmFEvhrnRoAEKlFVXVZLJwADL+O3ttIIdo1Q3KAIo8EPAPhPsgOoSJYSGbkEjUYhQIoYCScGA9AowdrIACMAAZ0eFYRAlj1ShgZrC4CCcnAigBHb7wNaRLIgMh0DRQOBwASFcJSKnsKRrMgbLZIXb-J5LWTsJotUXvOBfH7kpDNVrhGBkEiIvjI5AAJhVpXYfig+AAwhBZJsZhRCEUPHBKGqssKAYyYH8AbowG6nhAeDAML7tV6AL5BoA).

ts/types/ember-cli-blueprint-test-helpers/index.d.ts
chriskrycho: Looks like, weirdly, Prettier prefers to wrap the params across multiple lines if the return type is wrapped across multiple lines. Quirky!

ts/tests/blueprints/ember-cli-typescript-test.ts
chriskrycho: Turns out if you manually wrap the string to start on the next line, it does what we would all call the right thing.

.eslintrc.js
chriskrycho: Yeah, it turns out that including the plugin without setting any specific rules is what makes `yarn lint:s --fix` work for prettier fixes.

dfreeman: TIL

ts/tests/blueprints/ember-cli-typescript-test.ts
jamescdavis: Ah, very nice. Can you manually wrap this one?

chriskrycho: Yep, and in the latest change (1e6a85b) *is* wrapped. 🎉

jamescdavis: Wait, it's still not wrapped on this line: https://github.com/typed-ember/ember-cli-typescript/pull/678/files#diff-1411d46f2ad90f63434bf68ee3f6fcdeR141

chriskrycho: That's the old commit from when you made the original comment, I believe.

PR #684, opened 2019-04-16
Use 80-character line limit with Prettier.

chriskrycho:
We can do with this as we want; I just opened it so we could compare and see whether this is something we *want* or not.

In my own code style, I tend to put a soft limit at 80, including a hard limit for comments there, and hard limit at 100, but with Prettier I'm leaning increasingly toward the 80-character limit they recommend, because (as their docs note), it *maximizes* line length in a way that humans don't tend to.


dfreeman:
I think in probably 3/4 of the diff here, I like the single-line version better than the multi-line one, so I have a slight preference toward the 100-char limit, but ultimately I don't feel too strongly and am happy to go whichever way y'all like.

chriskrycho:
Alternative: 90-character limit with 3-space indentation. :trollface: 

I'm closing this because I don't think there's *enough* reason to switch, and it's needless churn in the codebase; we'll be fine with 100.

PR #728, opened 2019-05-23
chore: rename "toolchain" to "volta" in package.json

dfreeman:
Seems like it's probably time to switch this over?


dfreeman:
The Yarn registry seems to be having a rough day... running `yarn` just to refresh deps works well enough, but for a totally clean install the law of large numbers is kicking in.

I've now run `rm -r node_modules && yarn cache clean && yarn` a half dozen times and gotten a 503 error for a different package each time.

PR #749, opened 2019-07-03
Enable decorators in tsconfig blueprint

dustinsoftware:


chriskrycho:
Thanks again for opening this! Mind rebasing sometime in the next day or so? Our blocker from Ember CLI is now fixed, so it should work correctly at this point.

dustinsoftware:
@chriskrycho build passing now

PR #764, opened 2019-07-17
Stop installing old test helper declarations in the default blueprint

dfreeman:
We get frequent questions about the difference between `@types/ember-testing-helpers`, `@types/ember-test-helpers` and `@types/ember__test-helpers`. Since RFCs 232 and 268 (which became the default testing approach around Ember 3.0), we should only need the last one by default.


chriskrycho:
Ugh, rebased b/c of our "must be up to date with master" rule.

The main pain point that's solving given our velocity is for Renovate. But the bot rebases everything anyway. 🤔

PR #824, opened 2019-08-26
chore(deps): update ember-cli-addon-docs to v0.6.14

dfreeman:
The latest addon-docs release has looser dependency ranges for high-contention packages like `ember-concurrency` and `liquid-fire`, which (thanks to @jamescdavis's sleuthing) should get our CI green again.


jamescdavis:
Thanks, @dfreeman!

PR #857, opened 2019-09-12
refactor: convert addon build code to native class syntax

buschtoens:
As per [my message on Discord](https://discordapp.com/channels/480462759797063690/484421406659182603/621671315950272523) it seems that native class syntax is actually properly supported by `CoreObject` and `ember-cli`.


chriskrycho:
@buschtoens any chance you can make the suggested tweaks here?

buschtoens:
@chriskrycho Happy to do so, I just wasn't sure whether [`requireLocal`](https://github.com/typed-ember/ember-cli-typescript/pull/857#issuecomment-530864689) would be a "good enough" solution already.

chriskrycho:
We chatted about this, and we still *super* appreciate the work (three years later 😂) but are going to close it:

- We're working on an RFC to switch ember-cli-typescript into maintenance mode—details forthcoming, but the short version is: the broader ecosystem has moved around us to a default of having folks just run `tsc` directly as part of dev and CI, while using tools like Babel, ESBuild, SWC, etc. for the transpilation, and keeping those fully decoupled. We want to match that (and as folks move into an Embroider-native world, they will be able to opt into things like ts-loader in webpack if they *want* errors injected into the build).

- All the points discussed above back when this was opened originally.

Net, we want to minimize further direct investment in ember-cli-typescript internals. Thanks again, @buschtoens!

PR #866, opened 2019-09-19
chore(style): remove trailing space in the global.d.ts blueprint

scottkidder:
I found this trailing space in my `types/global.d.ts` when installing ember-cli-typescript.


ts/tests/blueprints/ember-cli-typescript-test.ts
```diff
@@ -136,7 +136,7 @@ describe('Acceptance: ember-cli-typescript generator', function() {
 
     const globalTypes = file('types/global.d.ts');
     expect(globalTypes).to.exist;
-    expect(globalTypes).to.include("declare module 'my-addon/templates/*'").to.include(`
+    expect(globalTypes).to.include("declare module 'my-addon/templates/*' {\n").to.include(`
```
chriskrycho: I *think* we can drop this particular addition to the test – note that we're not testing that there's a `}` to close it, for example.

scottkidder: Yeah I wasn't sure, it just felt right to start with a failing test.

PR #914, opened 2019-10-25
Support nullish coalescing and optional chaining with TS 3.7

dfreeman:
Brought to you by Project Night at Boston Ember

This PR updates ember-cli-typescript to add the Babel plugins for nullish coalescing and optional chaining if it detects a new enough project TypeScript version.

Until we update to TS 3.7 in this repo, though, I'm not 100% sure of a good way to test this (short of moving to a fancier multi-package workspaces-style testing setup, which we may eventually want to do anyway, but seems like a lot for this particular change). Definitely open to thoughts there.

I'm also assuming CI will fail until we get our Windows woes sorted 😢

### DIFF
--- a/package.json
+++ b/package.json
@@ -39,6 +39,8 @@
     "changelog": "conventional-changelog -p angular -i CHANGELOG.md -s -u"
   },
   "dependencies": {
+    "@babel/plugin-proposal-nullish-coalescing-operator": "^7.4.4",
+    "@babel/plugin-proposal-optional-chaining": "^7.6.0",
     "@babel/plugin-transform-typescript": "~7.6.0",
     "ansi-to-html": "^0.6.6",
     "debug": "^4.0.0",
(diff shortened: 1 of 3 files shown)


chriskrycho:
I'll test it with the 3.7 RC, and we should be fine to bump to 3.7 when it's out given the specific way we're using TS in the build pipeline, right?

dfreeman:
Yep. I suppose we _could_ even bump to the 3.7 rc now, since we have explicit CI coverage for the current release version as well.

chriskrycho:
Yeah, that makes sense as well. I'd be okay deferring this till it's officially out Just In Case™, but I don't have any strong objection to using the RC on `master` and just waiting to _publish_, either.

chriskrycho:
@dfreeman can you rebased this, and we can land it on master for people to experiment with? :pray:

dfreeman:
Got a green one!

PR #935, opened 2019-11-06
docs: Octane and much polish and clarification

chriskrycho:
This now migrates everything to GitBook, and covers both Octane and Classic. This is not exhaustive, but it *is* a reasonable starting point, and we can iterate from here.


tests/dummy/app/templates/docs/glimmer-components.md
```diff
@@ -0,0 +1,109 @@
...
+  }
+  
+  @action
+  minus() {
+    this.count -= 1;
+  }
+}
+```
+
+Notice that there are no type declarations here – but this *is* actually a well-typed component. The type of `count` is `number`, and if we accidentally wrote something like `this.count = null` the compiler would give us an error.
```
mike-north: > if we accidentally wrote something like `this.count = null` the compiler would give us an error.

only if `strictNullChecks` is turned on

tests/dummy/app/templates/docs/glimmer-components.md
```diff
@@ -0,0 +1,109 @@
...
+If you’re used to the classic Ember Object model, there are two important differences in the constructor itself:
+
+- we use `super` instead of `this._super`
+- we *must* call `super` before we do anything else
+
+</aside>
+
+Notice that we have to start by calling `super` with `owner` and `args`. This may be a bit different from what you’re used to in Ember or other frameworks, but is normal for sub-classes in TypeScript. If the compiler just accepted any `...arguments`, a lot of potentially *very* unsafe invocations would go through. So, instead of using `...arguments`, we explicitly pass the *specific* arguments and make sure their types match up with what the super-class expects.
+
+The types for `owner` here and `args` line up with what the `constructor` for Glimmer components expect. The `owner` is specified as `unknown` because this is a detail we explicitly *don’t* need to know about. The `args` are `object` because a Glimmer component *always* receives an object containing its arguments, even if the caller didn’t pass anything: then it would just be an empty object.
```
mike-north: > The `args` are `object`

for consistency, I would update the constructor signature to `args: object` to ensure readers aren't burdened with thinking about whether the empty interface type `{}` and the type `object` are equivalent (I myself had to check).

tests/dummy/app/templates/docs/glimmer-components.md
```diff
@@ -0,0 +1,109 @@
...
+
+The `args` are also available on `this`, just as you’d expect, so we could change our definition to return the names of the arguments from a getter:
```
mike-north: >  just as you’d expect

I'm not 100% sure that users will "just expect" this, given that if they've been working with ember components they have probably been using `this.foo` or the now deprecated `this.attrs.foo`. Do you think that readers will be better-served by treating `this.args` as a potentially new concept?

tests/dummy/app/templates/docs/glimmer-components.md
```diff
@@ -0,0 +1,109 @@
...
+<aside>
+
+If you’re used to the classic Ember Object model, there are two important differences in the constructor itself:
+
+- we use `super` instead of `this._super`
+- we *must* call `super` before we do anything else
+
+</aside>
+
+Notice that we have to start by calling `super` with `owner` and `args`. This may be a bit different from what you’re used to in Ember or other frameworks, but is normal for sub-classes in TypeScript. If the compiler just accepted any `...arguments`, a lot of potentially *very* unsafe invocations would go through. So, instead of using `...arguments`, we explicitly pass the *specific* arguments and make sure their types match up with what the super-class expects.
```
mike-north: My understanding is, that `...arguments` is only unsafe today because TS (so far) lacks support for [variadic kinds](https://github.com/Microsoft/TypeScript/issues/5453) which would give us things like typed `...rest` param tuples (https://github.com/microsoft/TypeScript/issues/5331)

If this is the case, we may want to list the known limitations, rather than describing broad and non-specific unsafe-ness (since it may one day become safe)

tests/dummy/app/templates/docs/glimmer-components.md
```diff
@@ -0,0 +1,109 @@
...
+  }
+}
+```
+
+<aside>
+
+If you’re used to the classic Ember Object model, there are two important differences in the constructor itself:
+
+- we use `super` instead of `this._super`
+- we *must* call `super` before we do anything else
```
mike-north: it might be useful to link to the part of the ECMA spec that states this as a hard requirement (I believe it is). 

This is the closest I could find
https://tc39.es/ecma262/#sec-runtime-semantics-classdefinitionevaluation

tests/dummy/app/templates/docs/glimmer-components.md
chriskrycho: Yeah, fair. I'll add a parenthetical somewhere explaining what the assumptions of the docs are. `strictNullChecks: true` is our *default*, so people have to explicitly opt out of it, and I'd like to keep these docs focused on the main path (while filling out answers and edge cases and so on in ancillary docs)—I don't think we should be trying to teach *TypeScript* with them, though given our audience I could see an argument that way. Along those lines, we might want a minimal TS primer that explains our configs and why we've chosen them. I'll add it to the project.

tests/dummy/app/templates/docs/glimmer-components.md
chriskrycho: I need to change the comment here to match the original, actually – I had updated the definition to match what the types on `Component` specify precisely to *avoid* explaining that. What this makes me realize is we probably need an `aside` linking to the explanation of the difference of the two.

tests/dummy/app/templates/docs/glimmer-components.md
chriskrycho: I'm assuming here and throughout (per the `aside` at the top of the doc) that this is a Glimmer Components With TypeScript guide, *not* a Glimmer Components In General guide. I'd really prefer to avoid repeating the materials in the guides—but I think spots like this might be a good place to put in further links in the "Is this bit new to you? Check out *this* part of the Ember Guides" vein. I'll add something like that here and push it up.

tests/dummy/app/templates/docs/glimmer-components.md
chriskrycho: Good point.

tests/dummy/app/templates/docs/glimmer-components.md
chriskrycho: Excellent suggestion, will do.

tests/dummy/app/templates/docs/glimmer-components.md
mike-north: Do we want this documentation to only apply for apps that preserve our generated config, or is there some benefit to making it config-agnostic where possible?

chriskrycho: Tradeoffs abound. I think we can probably solve this particular one by using an example that works in both settings (assign a string instead, for example). Maybe we make good use of that kind of example and then to *also* document our norms and why, and call it out if any particular point in the docs depends on our configs.

tests/dummy/app/templates/docs/glimmer-components.md
mike-north: Sounds like you're on top of this. I also would like to see us avoid duplicate documentation. Maybe something like 
```suggestion
The `args` passed to a Glimmer Component [are available on `this`](https://github.com/glimmerjs/glimmer.js/blob/2f840309f013898289af605abffe7aee7acc6ed5/packages/%40glimmer/component/src/component.ts#L12), so we could change our definition to return the names of the arguments from a getter:
```

chriskrycho: I've committed that for now, though I may opt to revise to a link to the Octane guides/API docs rather than the implementation in a later pass!

chriskrycho:
@dfreeman not sure why but we doesn't appear to be deploying this correctly (or possibly we have a caching issue). I can try to take a peek, but probably not till Monday. It's reporting latest published as 079a848 – when you merged master into this branch – and not picking up anything newer. 🤔

dfreeman:
🤦‍♂ The commits [are all there](https://github.com/typed-ember/ember-cli-typescript/commits/gh-pages), but apparently there's a known issue where _specifically for public repos_ a commit coming from an action won't trigger a Pages build: https://github.community/t5/GitHub-Actions/Github-action-not-triggering-gh-pages-upon-push/td-p/26869

I'll see if I can come up with a workaround, but for now running `ember deploy production` locally when you want to refresh may be the simplest path forward.

chriskrycho:
Whaaaaaaaat.

Deploy manually it is!

chriskrycho:
I've updated and added a bunch of material and redeployed! (I have not yet integrated the suggestions above b/c my focus was on getting a bunch of the organization in place.)

jamescdavis:
Apparently, you can [use an SSH key](https://github.community/t5/GitHub-Actions/Github-action-not-triggering-gh-pages-upon-push/m-p/34981#M2065) as a workaround.

tests/dummy/app/templates/docs/glimmer-components.md
chriskrycho: Resolved by 42fe7566.

tests/dummy/app/templates/docs/glimmer-components.md
chriskrycho: Curiously, that doesn't seem to actually specify it. All the implementations throw a reference error, treating `this` as an uninitialized variable—i.e. it's a Temporal Dead Zone—but I can't find *any* reference to this in spec or other official docs.

chriskrycho: Added a note in b7e4e99; I don't love it but it's better than nothing.

chriskrycho: Resolved by 42fe7566.

tests/dummy/app/snippets/cart-contents-lookup.ts
```diff
@@ -0,0 +1,18 @@
+// BEGIN-SNIPPET cart-contents-lookup.ts
+import Component from '@glimmer/component';
+import { getOwner } from '@ember/application';
+import { action } from '@ember/object';
+
+import ShoppingCartService from 'my-app/services/shopping-cart';
+
+export default class CartContentsComponent extends Component {
+  get cart() {
+    return getOwner(this).lookup('service:shopping-cart') as ShoppingCartService;
```
chriskrycho: You mentioned it in the writeup, but let's be *extra* explicit that this cast is wildly unsafe and the user has to guarantee it. If they *typo* the lookup, they'll get back `null` or something else. (It occurs to me that we could improve this by defining an `OwnerRegistry` and registering things there; mostly I wish that the `lookup` API were `lookup(bucket, name)` instead of `lookup('<bucket>:<name>')`, because if it were we could trivially reuse the existing registries.

jamescdavis: Definitely would be good to be extra explicit here. I was thinking the exact same thing about a registry for this.

chriskrycho: I opened [an issue on the RFCs repo](https://github.com/emberjs/rfcs/issues/583) for that idea, and will hopefully 🤞 write an actual RFC for it if there's interest.

tests/dummy/app/snippets/cart-contents.ts
```diff
@@ -0,0 +1,16 @@
+// BEGIN-SNIPPET cart-contents.ts
+import Component from '@glimmer/component';
+import { inject as service } from '@ember/service';
+import { action } from '@ember/object';
+
+import ShoppingCartService from 'my-app/services/shopping-cart';
+
+export default class CartContentsComponent extends Component {
+  @service shoppingCart: ShoppingCartService;
```
chriskrycho:
```suggestion
  @service shoppingCart!: ShoppingCartService;
```

tests/dummy/app/snippets/cart-contents-bad.ts
```diff
@@ -0,0 +1,17 @@
+// BEGIN-SNIPPET cart-contents-bad.ts
+import Component from '@glimmer/component';
+import { inject as service } from '@ember/service';
+import { action } from '@ember/object';
+
+import ShoppingCartService from 'my-app/services/shopping-cart';
+
+export default class CartContentsComponent extends Component {
+  @service shoppingCart: ShoppingCartService;
```
chriskrycho:
```suggestion
  @service shoppingCart!: ShoppingCartService;
```

tests/dummy/app/snippets/cart-contents.ts
jamescdavis: I'll also explain this.

tests/dummy/app/snippets/cart-contents-lookup.ts
chriskrycho: [RFC #585](https://github.com/emberjs/rfcs/pull/585) for the above concern!

dfreeman:
I haven't verified this myself yet, but there are reports that commits using the GH workflow token now correctly trigger a Pages publish for public repos: https://github.community/t5/GitHub-Actions/Github-action-not-triggering-gh-pages-upon-push/m-p/46527/highlight/true#M6555

chriskrycho:
@dfreeman @jamescdavis at this point, I think there's more value in getting these into the world than in thinking about trying to polish further first. They're already a year late. Better to get them up and then iterate on *something* and that way better serve end users. Assuming no 👎🏼 from either of you, I'll plan to merge this over the weekend and then we can keep making forward progress. We'll also need to coordinate getting the custom domain rejiggered for it, but that *should* be pretty easy.

jamescdavis:
Agreed! LMK when we're ready to cut over and I'll repoint the domain.

PR #946, opened 2019-11-11
fix: ensure we register Babel plugins with their full path

dfreeman:
Currently we're specifying just the name of the Babel plugins we register, which works as long as some version of those plugins gets hoisted and is available to `require` from the host application's perspective. This fails, though, if the hoisting doesn't happen, or if the package is linked, or other dependency shenanigans are in play.

This PR adds our plugins by their full path rather than just by name (the helpers we use already account for either of these formats, so nothing else should need to change).


chriskrycho:
Again: disabling Windows tests and merging. 😢

chriskrycho:
I have now reenabled Windows tests for the time being.

PR #955, opened 2019-11-14
Set up CI in GitHub Actions

dfreeman:
This PR introduces a definition for a "CI" GitHub workflow that aims to give equivalent coverage to that of our current Azure Pipelines setup.

Once we're happy with the setup, and assuming a quorum of the Typed Ember folks are on board, we can remove the Azure config files and update our Required Checks for PRs.

One note: I played some with `actions/cache` to persist dependencies across builds and speed things up a bit, but I'm not sure it's quite baked enough for our needs yet. It should be easy enough to layer on in the future if we figure out a way to use it effectively.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -1,0 +1,127 @@
+name: CI
+
+# These trigger patterns based on https://github.com/broccolijs/broccoli/pull/436
+on:
+  pull_request:
+  push:
+    # filtering branches here prevents duplicate builds from pull_request and push
+    branches:
+      - master
+      - 'v*'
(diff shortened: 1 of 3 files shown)


.github/workflows/ci.yml
```diff
@@ -0,0 +1,127 @@
...
+    branches:
+      - master
+      - 'v*'
+    # always run CI for tags
+    tags:
+      - '*'
+
+  # early issue detection: run CI weekly on Sundays
+  schedule:
+  - cron:  '0 6 * * 0'
```
chriskrycho: ❤️

PR #973, opened 2019-11-28
Fix precompile step

NullVoxPopuli:
Without this change I get this error when attempting to publish.:
```
Command failed with exit code 2: yarn tsc --allowJs false --noEmit false --rootDir /home/lprestonsegoiii/Development/NullVoxPopuli/ember-contextual-services --isolatedModules false --declaration --declarationDir /tmp/e-c-ts-precompile-18769 --emitDeclarationOnly
```

Only after this change am I able to publish addons. 

:man_shrugging:

PR #984, opened 2019-12-02
Ensure errors are emitted when ts:precompile fails

dfreeman:
As noted both in Discord and in #973, we currently don't emit any of `tsc`'s output when it fails during `ts:precompile`. This ensures we capture and re-emit any output when the command exits with a nonzero code, and adds a test to make sure we don't regress in the future.


dfreeman:
Floating-deps tests should start passing again once https://github.com/zloirock/core-js/issues/710 is resolved

chriskrycho:
Updated from master, rerunning CI now. Also reviewing!

ts/lib/commands/precompile.ts
```diff
@@ -25,18 +25,25 @@ export default command({
...
+        '--noEmit', 'false',
+        '--rootDir', rootDir || this.project.root,
+        '--isolatedModules', 'false',
+        '--declaration',
+        '--declarationDir', outDir,
+        '--emitDeclarationOnly',
+        '--pretty', 'true',
+      ], {
+        preferLocal: true,
+        all: true,
```
chriskrycho: Would you mind adding a comment here explaining why we want this? Will make maintenance easier. (I just went and read the `execa` docs to figure it out, figure it'll save us time in the future from doing the same.)

dfreeman: Done! `execa` is one of those that I've used enough places that I've lost any good sense of which options are self-describing and which aren't 🙃

chriskrycho: I sympathize!

ef4:
Thanks, I just hit this bug and was about to make the same PR when I saw it's already fixed on master.

jamescdavis:
Released in [3.1.2](https://github.com/typed-ember/ember-cli-typescript/releases/tag/v3.1.2)

PR #993, opened 2019-12-07
Docs custom domain

jamescdavis:
Turns out all we need to do here is set the root URL to ''.

After merging this PR, the steps for enabling the custom domain are:
1. re-deploy `master` (and update latest):
`ADDON_DOCS_UPDATE_LATEST=true ember deploy production`
2. immediately enable the custom domain [here](https://github.com/typed-ember/ember-cli-typescript/settings#pages-cname-field) by entering:
`ember-cli-typescript.com`
3. update and re-deploy the `octane-docs` branch so it continues to work too:
`git checkout octane-docs && git merge master && ember deploy production`

I've tested this on my fork and then removed the custom domain there so typed-ember can pick it up. DNS is already configured for github pages.

Once re-deployed & configured, docs will be available at https://ember-cli-typescript.com and https://typed-ember.github.io/ember-cli-typescript/ will redirect there.

Also, ember-cli-addon-docs has a bug, which has been fixed with https://github.com/ember-learn/ember-cli-addon-docs/pull/412, but not released yet, so we need to pin to a sha to re-deploy and have our blue brand customization work. We're [already doing this](https://github.com/typed-ember/ember-cli-typescript/blob/b9f3c7bb3d053c985faca742fc8c5d0433e8281c/package.json#L87) in the `octane-docs` branch. This will just bring it into `master` so we can re-deploy `master` too.

### DIFF
--- a/config/addon-docs.js
+++ b/config/addon-docs.js
@@ -6,4 +6,7 @@ const AddonDocsConfig = require('ember-cli-addon-docs/lib/config');
 module.exports = class extends AddonDocsConfig {
   // See https://ember-learn.github.io/ember-cli-addon-docs/docs/deploying
   // for details on configuration you can override here.
+  getRootURL() {
+    return '';
+  }
 };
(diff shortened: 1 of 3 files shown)


chriskrycho:
💥

PR #995, opened 2019-12-08
chore(deps): force update of terser-webpack-plugin so we get serialize-javascript@2.1.1

jamescdavis:
Removed `terser-webpack-plugin` entry from `yarn.lock` and re-yarned to force update to latest indirect dependencies.

Dep tree:

```
ember-auto-import ->
  webpack ->
    terser-webpack-plugin ->
      serialize-javascript
```

https://github.com/webpack-contrib/terser-webpack-plugin/releases/tag/v1.4.2


chriskrycho:
Thanks not only for the change but the helpful explanatory comment!

PR #1018, opened 2019-12-17
[v4] allow ember-cli-babel to manage TS transpilation

dfreeman:
Based on [a conversation with some typed-ember folks and @rwjblue](https://discordapp.com/channels/480462759797063690/484421406659182603/654770983319896064), our likely plan moving forward is to have `ember-cli-babel` be responsible for setting up any relevant TypeScript Babel plugin(s) using the presence of `ember-cli-typescript` as a trigger. The rough sequence of events I'm picturing:
 - a `4.0.0-alpha.1` release of this package that includes this PR to stop touching Babel config on this side
 - a release of `ember-cli-babel` that looks for `ember-cli-typescript >= 4.0.0-alpha` and sets up the appropriate plugins if it finds them
 - subsequent updates here to enforce that we're always installed alongside at least that ember-cli-babel version
 - ...
 - `ember-cli-typescript@4.0.0` 🎉 

Note that this PR is targeting the v4 branch, which we can use as a central point to cut alpha releases and land other changes we want to include in our next major release as we coordinate with ember-cli-babel on the rollout.

### DIFF
--- a/package.json
+++ b/package.json
@@ -118,7 +118,7 @@
     "in-repo-b": "link:tests/dummy/lib/in-repo-b",
     "loader.js": "4.7.0",
     "mocha": "6.2.2",
-    "prettier": "1.18.2",
+    "prettier": "1.19.1",
     "qunit-dom": "0.9.2",
     "testdouble": "3.12.4",
     "tmp": "0.1.0",
--- a/package.json
+++ b/package.json
@@ -39,12 +39,8 @@
     "changelog": "conventional-changelog -p angular -i CHANGELOG.md -s -u"
   },
   "dependencies": {
-    "@babel/plugin-proposal-nullish-coalescing-operator": "^7.4.4",
-    "@babel/plugin-proposal-optional-chaining": "^7.6.0",
-    "@babel/plugin-transform-typescript": "~7.7.0",
     "ansi-to-html": "^0.6.6",
     "debug": "^4.0.0",
-    "ember-cli-babel-plugin-helpers": "^1.0.0",
     "execa": "^3.0.0",
     "fs-extra": "^8.0.0",
     "resolve": "^1.5.0",
@@ -140,8 +136,7 @@
   "ember-addon": {
     "configPath": "tests/dummy/config",
     "before": [
-      "broccoli-watcher",
-      "ember-cli-babel"
+      "broccoli-watcher"
     ]
   },
   "husky": {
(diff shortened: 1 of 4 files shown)


dfreeman:
Note: it's expected (and unavoidable) that CI will fail here until a version of `ember-cli-babel` exists that does what the code I just deleted would have done

package.json
```diff
@@ -39,12 +39,8 @@
     "changelog": "conventional-changelog -p angular -i CHANGELOG.md -s -u"
   },
   "dependencies": {
-    "@babel/plugin-proposal-nullish-coalescing-operator": "^7.4.4",
-    "@babel/plugin-proposal-optional-chaining": "^7.6.0",
-    "@babel/plugin-transform-typescript": "~7.7.0",
```
chriskrycho: 1. To make this work, will we need ember-cli-babel to introspect against not only this but the `typescript` version installed to make sure things are set up correctly?
2. Don't *we* need these internally for our *own* compilation pipeline? Or is the idea that our dependency on `ember-cli-babel` will handle it for us as well? (I'm trying to think through the circularity here.)

jamescdavis:
I've opened a draft PR for the `ember-cli-babel` side of this: https://github.com/babel/ember-cli-babel/pull/314

package.json
jamescdavis: 1. That was my understanding and what I did in https://github.com/babel/ember-cli-babel/pull/314
2. We really don't have our *own* compilation pipeline (well, we do, but it doesn't use Babel).

dfreeman: 1. For the same reason we [couldn't do it here](https://github.com/typed-ember/ember-cli-typescript/pull/936), we unfortunately can't conditionalize plugin inclusion on our parent's TS version in ember-cli-babel. Tl;dr: for anything other than the root app, we don't know what version of TypeScript the package was running against; at best we just have a semver specifier.

2. What @jamescdavis said 🙂 — our build files are transpiled via `tsc`, and all our runtime test code will be handled by ember-cli-babel once it's updated.

chriskrycho: DOH. Thank you both for patiently clarifying for me. (Let’s just write this off as a symptom of my just having spent 3½ months upgrading a massive app which had a ton of private router API usage from 3.4 to 3.8. :sob:)

jamescdavis: 1. I was pretty sure that was the case, but I figured I'd go ahead and add the conditionality and we could remove it if it needed.

jamescdavis:
I've updated the tests. This now passes when yarn linked to https://github.com/babel/ember-cli-babel/pull/314 and version bumped to `4.0.0-alpha.0` (or greater).

PR #1026, opened 2019-12-20
docs: fix reference to Service `mySession`

MrChocolatine:
The Service is injected in the Class through the key `mySession` but is referred to `session` in the `login()` method.


chriskrycho:
Thanks!

PR #1044, opened 2020-01-18
Use Volta for CI instead of setup-node action

jamescdavis:
This configures the CI workflow to use [Volta](https://volta.sh/) instead of the setup-node action.


jamescdavis:
Nevermind
> Run rwjblue/setup-volta@v1
downloading volta@0.7.0
##[error]windows is not yet supported
##[error]Node run failed with exit code 1

Issue #780, opened 2019-07-31
Docs: add info about merged app directory


chriskrycho:
We [currently document][current] one gotcha for addon authors *building* addons in TypeScript:

> Because addons have no control over how files in `app/` are transpiled, **you cannot have `.ts` files in your addon's `app/` folder.**

[current]: https://typed-ember.github.io/ember-cli-typescript/versions/master/docs/ts-guide/with-addons#key-differences-from-apps

To expand on that for future readers who may stumble on this (and as a starting point for the docs we should write up):

Having `.js` supplied to a specific path inside the `app` namespace by both an addon and the app itself, when one of them is using TypeScript to generate that file, creates a caching problem. When addons inject `.js` files into the `app` namespace, *and* we're compiling `.ts` files to `.js` files in the consuming app (*always* in the `app` namespace), or vice versa, we can end up in a situation where Ember CLI sees that there is already a `.js` file there, so it ignores the one generated from `.ts`. Which one it picks up seems to be semi-random (at least: we haven't chased it down to see if it's consistently replicable).

The *real* solution here is for addons to stop injecting themselves into the app namespace and to supply imports people can use instead to explicitly opt into using their functionality. However, until that becomes a community norm (best guess: not until Embroider lands), we should document this as a gotcha, including the kinds of build errors it produces, so that people find it when it happens to them and we have a useful place to direct addon authors to in understanding why and how they're breaking their consumers.

## Ecosystem issues

- #724
- ember-cli/ember-ajax#428
- stefanpenner/broccoli-persistent-filter#166

lifeart:
// cc

I think it should be `redlined` and pinned in the docs - because this is floating behaviour and can create serious damage for business.

for case - we have `application.ts` route and have `ember-simple-auth` installed - it's so easy to get issue with inconsistent builds, because there is no user action needed for it. (and alot of ember users use `ember-simple-auth`)

jordpo:
I've also gotten bitten by the ESA Application route class export stomping my TS class.

LuisAverhoff:
I've also experienced this same issue with `ember-simple-auth` stomping my TS class.

patrickberkeley:
This is a big trap. Any app that consumes an addon that exports a file that matches the name of a file in the app will sporadically get a bad build.

ESA is one of the most popular add-ons, and it happens to export the application route. So that is why people have come across this issue using it.

[...]

Issue #724, opened 2019-05-16
Inconsistent builds with addons having files with same names


lifeart:
reproduction repo: https://github.com/lifeart/ts-merging-issue

how to reproduce?
`yarn; ember test; ember test; ember test` 
or see

```
WARNING: ts-app has added the class-properties plugin to its build, but ember-cli-babel provides these by default now! You can remove the transforms, or the addon that provided them, such as @ember-decorators/babel-transforms.
WARNING: ts-app has added the class-properties plugin to its build, but ember-cli-babel provides these by default now! You can remove the transforms, or the addon that provided them, such as @ember-decorators/babel-transforms.
Environment: test
cleaning up...
Built project successfully. Stored in "C:\Users\lifeart\Documents\repos\tmp\ts-app\tmp\class-tests_dist-UBMhIWw3.tmp".
ok 1 Chrome 74.0 - [2 ms] - ESLint | app: app/models/meta-model.js
ok 2 Chrome 74.0 - [0 ms] - ESLint | app: app.js
ok 3 Chrome 74.0 - [1 ms] - ESLint | app: resolver.js
ok 4 Chrome 74.0 - [0 ms] - ESLint | app: router.js
ok 5 Chrome 74.0 - [0 ms] - TemplateLint: ts-app/templates/application.hbs
ok 6 Chrome 74.0 - [0 ms] - ESLint | tests: test-helper.js
ok 7 Chrome 74.0 - [108 ms] - Unit | Model | base model: it exists
ok 8 Chrome 74.0 - [25 ms] - Unit | Model | meta model: it exists
ok 9 Chrome 74.0 - [0 ms] - ember-qunit: Ember.onerror validation: Ember.onerror is functioning properly

1..9
# tests 9
# pass  9
# skip  0
# fail  0

# ok
WARNING: ts-app has added the class-properties plugin to its build, but ember-cli-babel provides these by default now! You can remove the transforms, or the addon that provided them, such as @ember-decorators/babel-transforms.
WARNING: ts-app has added the class-properties plugin to its build, but ember-cli-babel provides these by default now! You can remove the transforms, or the addon that provided them, such as @ember-decorators/babel-transforms.
Environment: test
cleaning up...
Built project successfully. Stored in "C:\Users\lifeart\Documents\repos\tmp\ts-app\tmp\class-tests_dist-s6n8AZfw.tmp".
ok 1 Chrome 74.0 - [2 ms] - ESLint | app: app/models/meta-model.js
[...]

PR #1046, opened 2020-01-20
Issue a warning if we detect a .js/.ts file collision

dfreeman:
This PR will cause a warning to be emitted if we detect the situation described in #780. Because of exactly the situation we're detecting, we can't tell _where_ the two conflicting files are coming from, but in most cases it will be apparent from context (e.g. the host app overriding a service from a particular addon).

Trying to describe the full situation and its possible ramifications would produce a giant wall of text in the terminal output, but the link in the warning points to #780, which has some good context and also links out to additional issues, like the upstream one in `broccoli-persistent-filter`.

image


lifeart:
Issue happens if both files is `.ts` also

dfreeman:
> also issue happens if both files is .ts

I don't think this is correct. Addons should never have `.ts` files in their `app/` trees, but if they do and the host also has a `.ts` file, it will override the addon's file.

lifeart:
https://github.com/typed-ember/ember-cli-typescript/issues/724

dfreeman:
Nothing in that thread indicates an issue when there were two conflicting `.ts` files when there wasn’t also a `.js` file of the same name floating around (e.g. because of differing addon versions that switched from including a `.ts` file to including a `.js` one)

If you can provide a reproduction where a nondeterministic build results from _only_ `.ts` files I’m happy to take a look, but if you consider the underlying problem here, that _can’t_ be caused by the same root issue as the one this PR is detecting, so it would need to be investigated and remedied separately.

ts/addon.ts
```diff
@@ -97,6 +106,47 @@ export default addon({
...
+        if (path !== '/') return input;
+
+        let addon = this;
+        let checked = false;
+        let stew = require('broccoli-stew') as typeof import('broccoli-stew');
+
+        return stew.afterBuild(input, function() {
+          if (!checked) {
+            checked = true;
+            addon._checkForFileCollisions(this.inputPaths[0]);
```
chriskrycho: Just checking: what invariant guarantees that `this.inputPaths` always has at least one element?

ts/tests/helpers/skeleton-app.ts
```diff
@@ -44,7 +44,7 @@ export default class SkeletonApp {
     ));
   }
 
-  updatePackageJSON(callback: (arg: any) => string) {
+  updatePackageJSON(callback: (arg: any) => any) {
```
chriskrycho: 😭

dfreeman: Really this is `(arg: PackageJSON) => PackageJSON | void` — `string` was never right 🙃

ts/addon.ts
dfreeman: Broadly speaking, Broccoli plugins have one element in `inputPaths` per node they receive in their constructor.

In the case of `afterBuild` (and `Funnel` and a lot of other common plugins), it accepts a single node and then passes it in a singleton array when it calls `super` in its constructor, so `inputPaths[0]` in this case is the path on disk for the `input` tree we pass in a couple lines above.

chriskrycho: Excellent; thank you! TIL!

jamescdavis:
This has been released in [v3.1.3](https://github.com/typed-ember/ember-cli-typescript/releases/tag/v3.1.3).

azhiv:
Can this warning be treated as an error during build? It would be nice to fail the build in case of a collision because otherwise proceeding with the collision (if missed) may result in significant harm.

PR #1153, opened 2020-05-28
Use temp dir in project for precompile

jamescdavis:
This changes `ts:precompile` to use a temp dir within the project (under cwd) instead of a system temp file. This is to work around https://github.com/microsoft/TypeScript/issues/34749#issuecomment-549916145

It also makes sure the temp dir is removed after the command is run, even if an error is encountered.

### DIFF
--- a/ts/lib/commands/precompile.ts
+++ b/ts/lib/commands/precompile.ts
@@ -42,6 +42,7 @@ export default command({
         all: true,
       });
     } catch (e) {
+      fs.removeSync(outDir);
       console.error(`\n${e.all}\n`);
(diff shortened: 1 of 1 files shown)


jamescdavis:
I have not been able to repro either of those test failures locally ¯\\\_(ツ)\_/¯.

dfreeman:
Hmmm... maybe related to https://github.com/emberjs/ember.js/pull/18831? I haven't totally digested the solution (or even the problem) there yet, so I'm not   sure if it makes sense that we'd be running into that or not.

PR #1162, opened 2020-06-16
docs: fix description of ts:precompile

hjdivad:
Limit the description to only the bit that's still true, i.e. that it generates declaration files.


hjdivad:
@chriskrycho seems the linter has defeated my attempts to make this change from the website.

Maybe with a little `yarn lint --fix` CI can be happier.

chriskrycho:
Doh! Much obliged.

PR #1169, opened 2020-07-01
Bring apps closer to typechecking out of the box

dfreeman:
Today, if I run `ember new my-app` and then immediately follow that with `ember install ember-cli-typescript`, I can _almost_ rename `app.js`, `router.js` and `test-helper.js` to end in `.ts` and have a typechecking application out of the box. There are still just a couple of blockers for that:
 - `app.js` imports `ember-resolver` but we don't install the `@types` package for that by default
 - `test-helper.js` references `config.APP`, which isn't part of the declarations we emit for `config/environment`
 - `test-helper.js` includes relative imports for `app` and `config/environment` that work at runtime but don't match the on-disk layout, so TS rejects the import

This PR aims to fix those first two bullets, and I have [an ember-cli PR open](https://github.com/ember-cli/ember-cli/pull/9273) to address the third.


ts/blueprints/ember-cli-typescript/index.js
```diff
@@ -146,8 +146,8 @@ module.exports = {
       { name: 'ember-cli-typescript-blueprints', target: 'latest' },
       { name: 'typescript', target: 'latest' },
       { name: '@types/ember', target: 'latest' },
-      { name: '@types/rsvp', target: 'latest' },
```
dfreeman, line 149: `@types/ember` already brings in `@types/rsvp`, and having one fewer explicit version constraint floating around makes "cannot re-declare block scoped variables" more likely to crop up as folks update various dependencies.

PR #1191, opened 2020-07-02
Mass updates

dfreeman:
This gets the vast majority of our dependencies up to date, with a few changes (like updating `prettier`) pulled into their own commits for clarity. I'm hoping to get the Renovate traffic jam cleared out.


dfreeman:
Before merging this, we may want to create a `v3` branch off `master`. Several of these updates require Node 10, which is already what we're running CI against, but 3.x officially supports Node 8 still.

dfreeman:
For future travelers: we decided to create any potential `v3` branch based on our last 3.x tag rather than from `master` to be sure we don't accidentally include something that depends on Node 10 that got Renovated in in the meantime.

PR #1198, opened 2020-07-03
fix(typecheck-worker): avoid false-positive `willTypecheck`s on Linux

dfreeman:
We've seen occasional timeouts in CI for our test that verifies we don't block builds for file changes that don't trigger a new typecheck. Interestingly, these timeouts have never occurred on Windows, and I've also never managed to reproduce them locally on my Mac, but they pretty reliably occur every 3-4 runs on Linux.

Digging in (by running the offending test in a loop in CI), it seems that the change to a `.hbs` file is occasionally causing TypeScript's `createProgram` to be called, but then subsequently short circuited as some other piece of machinery discovers that the change in question doesn't matter. This is a problem for us, as we treat `createProgram` as proof that a typecheck has started, and if `afterProgramCreate` is never subsequently called, then our worker just sits on a pending promise forever.

Fortunately for us, when `afterProgramCreate` _is_ called, it's called synchronously near the end of `createProgram`, so from the perspective of the outside world, it doesn't matter if we call `willTypecheck` in `createProgram` or `afterProgramCreate`: no events will be dispatched in between those two functions anyway.

### DIFF
--- a/ts/lib/typechecking/worker/index.ts
+++ b/ts/lib/typechecking/worker/index.ts
@@ -138,7 +138,7 @@ export default class TypecheckWorker {
   private patchCompilerHostMethods(
     host: WatchCompilerHostOfConfigFile<SemanticDiagnosticsBuilderProgram>
   ) {
-    let { watchFile, watchDirectory, createProgram, afterProgramCreate = () => {} } = host;
+    let { watchFile, watchDirectory, afterProgramCreate = () => {} } = host;
 
     // Intercept tsc's `watchFile` to also invoke `mayTypecheck()` when a watched file changes
     host.watchFile = (path, callback, pollingInterval?) => {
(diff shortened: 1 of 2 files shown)


dfreeman:
P.S. I ran the offending test 50 times in a loop in CI and didn't get any more hangs with this change.

chriskrycho:
For clarification on my comment for any lookers-on/posterity: while I would guess this is fairly stable behavior, I inherently dislike solutions that leave us at the mercy of changes to an API that *could* change whenever convenient for the TS team—but since it's fairly core to the compiler API, it seems reasonable and we expect that it would make release notes/etc. if it *were* to change. And in truth, we're *always* in that position so…

jamescdavis:
Thank you, @dfreeman !

Issue #1272, opened 2020-08-21
Remove rootDir option from precompile command


NullVoxPopuli:
### Please paste the output of `ember -v` here

ember-cli: 3.16.x
node: 12.x
os: linux x64

### Please paste the output of `tsc -v` here

Version 3.9.7

### Please paste the version of `ember-cli-typescript` and `ember-cli-typescript-blueprints` here

```
$ yarn list --pattern 'ember-cli-typescript'
yarn list v1.22.4
warning Resolution field "d3-selection@1.3.2" is incompatible with requested version "d3-selection@1.3.0"
├─ @glimmer/component@1.0.0
│  └─ ember-cli-typescript@3.0.0
├─ ember-cli-typescript-blueprints@3.0.0
├─ ember-cli-typescript@4.0.0-rc.1                <------------ this is the one used throughout all projects in our monorepo
├─ ember-concurrency-decorators@1.0.0
│  └─ ember-cli-typescript@2.0.2
├─ ember-fetch@7.0.1
│  └─ ember-cli-typescript@3.1.4
├─ ember-lifecycle-component@0.7.0
│  └─ ember-cli-typescript@3.1.4
├─ ember-load-initializers@2.1.1
│  └─ ember-cli-typescript@2.0.2
├─ ember-pattern-librarian@1.1.7
│  └─ ember-test-waiters@2.1.1
│     └─ ember-cli-typescript@3.1.4
└─ tracked-built-ins@1.0.2
   └─ ember-cli-typescript@3.1.4
Done in 2.75s.
```

### Please paste your `tsconfig.json` and `tslint.json` or `eslint.json` (if applicable) below

Our tsconfig is in two parts:

root:

<details><summary><b>tsconfig.json at the root with shared options</b></summary>

```
// https://www.typescriptlang.org/v2/en/tsconfig
{
  "compilerOptions": {
    // Babel handles transpiling, not tsc
    "target": "ESNext",
    "module": "ESNext",

    // Many npm modules are not distributed as actual modules
    "allowSyntheticDefaultImports": true,

    // Some of these overlap with @typescript-eslint a bit, but we can fail faster
    // be also having the type checker check these.
    //
    // -- cleanliness
    "noUnusedLocals": true,
    "noUnusedParameters": true,
    "noImplicitReturns": false,
    // -- correctness
    "noImplicitAny": true,
    "noImplicitThis": true,
    "alwaysStrict": true,
    "strictNullChecks": true,
[...]

PR #1273, opened 2020-08-21
resolve GH#1272 by removing the rootDir option from the precompile co…

NullVoxPopuli:
Related issue: https://github.com/typed-ember/ember-cli-typescript/issues/1272

### DIFF
--- a/ts/lib/commands/precompile.ts
+++ b/ts/lib/commands/precompile.ts
@@ -15,7 +15,7 @@ export default command({
 
   async run(options: { manifestPath: string }) {
     let outDir = `${process.cwd()}/e-c-ts-precompile-${process.pid}`;
-    let { paths, rootDir, pathRoots } = this._loadConfig(outDir);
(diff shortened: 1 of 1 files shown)


dfreeman:
In the short term I don't think we can get away with not setting `rootDir` because of the impact it has on how generated declarations are structured.

In general, the `ts:precompile` command has been troublesome for folks because it _has_ to set certain flags to be compatible with `--declaration` and to be able to interpret the output correctly.

For that reason, I'd love to be eventually able to eliminate the `ts:precompile` command entirely in favor of a composite build-based setup where just running `tsc` emits declarations in the right place. There's still some work to be done to make that ergonomic in Ember projects, though, and as a prerequisite for that: making sure it _can_ be ergonomic before we commit to it!

chriskrycho:
Per Dan's comment and also the note on #1272 and our other plans to shift how we approach this, closing this. Thanks for the attempt at a fix here!

Issue #1341, opened 2020-10-09
@types/ember is not used for auto completion


mydea:
I have been wondering for a while why I was not getting any auto complete suggestions for any `@ember/...` packages in VSCode. After some playing around, I found out that it only seems to provide auto-complete for direct dependencies defined in the package.json file. Since all the ember types are actually provided as sub-dependencies via `@ember/types`, they are not used. After I replaced `@types/ember` with the full list from there:

```
{
        "@types/ember__application": "*",
        "@types/ember__array": "*",
        "@types/ember__component": "*",
        "@types/ember__controller": "*",
        "@types/ember__debug": "*",
        "@types/ember__engine": "*",
        "@types/ember__error": "*",
        "@types/ember__object": "*",
        "@types/ember__polyfills": "*",
        "@types/ember__routing": "*",
        "@types/ember__runloop": "*",
        "@types/ember__service": "*",
        "@types/ember__string": "*",
        "@types/ember__template": "*",
        "@types/ember__test": "*",
        "@types/ember__utils": "*",
        "@types/htmlbars-inline-precompile": "*",
        "@types/jquery": "*",
        "@types/rsvp": "*"
}
```

(Replacing the stars with actual versions)

Auto complete worked like a charm. For me as a non-TS-expect, that was rather unexpected.

* As a developer, I want to get autocompletion for ember packages out of the box after installing ember-cli-typescript

Would it make sense to either install these directly instead of `@types/ember`? It _is_ a bit sub-ideal that it kind of bloats up your package.json, but if this is what is needed for auto-complete I'll gladly take it.

chriskrycho:
Thanks for filing this! The problem is that TS *intentionally* does not suggest imports from transitive dependencies, since it's not safe to depend on them; but Ember supplies its packages… as transitive dependencies. (That's not even *really* what's happening today; it's still relying on build-time magic transforms of imports, but that's notionally what's happening, and presumably/hopefully what the future holds.) We need to update our blueprints to explicitly install the `@types/ember__*` dependencies as well as `@types/ember`, and the same for `@types/ember-data__*` as well as `@types/ember-data`.

mydea:
If you're willing to accept this, I'd be happy to update the blueprints accordingly!

chriskrycho:
Yes please!

PR #1383, opened 2020-11-13
feat: install @types/ember packages separately via blueprint

mydea:
This PR updates the install blueprint to install the `@types/ember__XXX` packages separately, instead of installing only `@types/ember`.

> The problem is that TS intentionally does not suggest imports from transitive dependencies, since it's not safe to depend on them;

Some notes:

* I did not install `@types/ember` itself - I have been running our app like this for a while, and I think you'd only need it if you use the Ember namespace (e.g. `import Ember from 'ember'; Ember.get`), which is not encouraged anymore anyhow and can/should be mostly avoided.
* For `@types/ember-data` I did not find any direct dependencies - not sure how exactly that works there with e.g. `@types/ember-data__model`, I guess you'd have to manually install those even now?

Closes https://github.com/typed-ember/ember-cli-typescript/issues/1341

### DIFF
--- a/ts/blueprints/ember-cli-typescript/index.js
+++ b/ts/blueprints/ember-cli-typescript/index.js
@@ -144,32 +144,53 @@ module.exports = {
     }
 
     let packages = [
-      { name: 'ember-cli-typescript-blueprints', target: 'latest' },
-      { name: 'typescript', target: 'latest' },
-      { name: '@types/ember', target: 'latest' },
-      { name: '@types/ember__test-helpers', target: 'latest' },
-      { name: '@types/ember-resolver', target: 'latest' },
+      'typescript',
+      'ember-cli-typescript-blueprints',
+      '@types/ember-resolver',
+      '@types/ember__test-helpers',
+      '@types/ember__object',
+      '@types/ember__service',
+      '@types/ember__controller',
+      '@types/ember__string',
+      '@types/ember__template',
+      '@types/ember__polyfills',
+      '@types/ember__utils',
+      '@types/ember__runloop',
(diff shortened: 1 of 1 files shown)


chriskrycho:
@mydea thank you so much for opening this!

@dfreeman I believe you most recently poked at this—my understanding is we probably *do* still want to install `@types/ember` directly as well, even though we’re not normally importing from the Ember namespace directly anymore, if for no other reason than that dependency management (esp. w/any transitives) is much easier if it’s direct. Does that sound right to you?

dfreeman:
I think the extent of my poking (unless I'm forgetting another conversation) was what @mydea included in the description—something along the lines of "oh look, auto-import works if the corresponding `@types/ember__` package is a direct dependency 🤔"

I _think_ the only reason to still include `@types/ember` is if you need to import from the `'ember'` module, which is mostly only necessary if you're using _very_ legacy APIs.

The one exception is the array prototype extensions, which it looks like we [still enable types for by default](https://github.com/typed-ember/ember-cli-typescript/blob/master/ts/blueprints/ember-cli-typescript/index.js#L9). Since the source of truth for those is in `@types/ember`, we likely either need to continue installing that by default or comment out that line in the blueprint and include a note that you need `@types/ember` for those to work.

chriskrycho:
Ah, okay, that makes sense. I think I'm fine with changing the default here, though we'll want to note it and make sure it gets hit in the docs (if/as/when we actually finish updating them 😭).

chriskrycho:
Landing this—we can iterate on the `@types/ember` question later!

PR #1391, opened 2020-12-08
docs: Update configuration.md to include an example of generating sourcemaps with Embroider

johanrd:
I had to add devtool to my webpackConfig in order for  sourcemaps to be generated of my typescript files with Embroider. 

Not sure if this documentation belongs here or in the embroider docs (or whether there are other/better ways to set up the config)


chriskrycho (approved):
This is great—thank you!

chriskrycho:
(Note: I updated the branch here so it passes our commit lint setup. I'm likely going to turn that off in the future, as it's not really adding value to us, esp. for things like this.)

PR #1401, opened 2021-02-03
Fix qunit types... again

jamescdavis:
This bumps `@types/qunit` and `yarn-deduplicate`s to resolve the failing floating dependency tests... again.


chriskrycho (approved):
And so it will forever. «sigh»

PR #1410, opened 2021-02-22
docs: add Ember Data transform page

SergeAstapov:
We started integrating TypeScript into existing application and stumbled upon custom transform.

https://docs.ember-cli-typescript.com/ember-data/models wasn't much helpful so I followed https://github.com/DefinitelyTyped/DefinitelyTyped/pull/50328/files and had to add interface to `TransformRegistry` to make TypeScript happy:

```typescript
declare module 'ember-data/types/registries/transform' {
  export default interface TransformRegistry {
    point: PointTransform;
  }
}
```

after that everything started to work as expected.

I know the prose is really bad but could not came up with anything better. Maybe we can collectively iterate over the prose if the snippets are good here.

### DIFF
--- a/docs/ember-data/transforms.md
+++ b/docs/ember-data/transforms.md
@@ -1,0 +1,53 @@
+# Transforms
+
+`DS.attr` defines an attribute on a [DS.Model](https://guides.emberjs.com/release/models/defining-models/).
+By default, attributes are passed through as-is, however you can specify an
+optional type to have the value automatically transformed.
+Ember Data ships with four basic transform types: `string`, `number`, `boolean` and `date`.
+
+You can define your own transforms by subclassing [DS.Transform](https://guides.emberjs.com/release/models/defining-models/#toc_custom-transforms).
+Ember Data transforms are normal TypeScript classes.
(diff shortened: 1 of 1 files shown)


SergeAstapov:
Is this actually right/recommended way of extending `TransformRegistry`? Should we instead recommend to define these via `types/registries/transform.d.ts` like
```
export default interface TransformRegistry {
  point: PointTransform;
}
```

I'm new to ember-cli-typescript so would love to be guinea pig and test existing docs and identify what is missing

chriskrycho:
Hi @SergeAstapov – thanks so much for this! I apologize I missed it when it came through; I need to tweak my notifications settings here. I'll try to review it on Friday!

SergeAstapov:
Thank you @chriskrycho! I update PR to use more complex example with custom data structure. Please advice if this is not the canonical/suggested way of doing transforms, would love to hear back.

chriskrycho:
@jamescdavis mind taking a look at this? You’re more up-to-date and well-informed on Ember Data than I am!

chriskrycho:
I just went ahead and merged this; can iterate later! Thanks @SergeAstapov!

PR #1415, opened 2021-03-17
chore: update option `noEmitOnError` to `true`

### LINKED ISSUE #1402: Docs on "Using TS Effectively" are wrong about blueprint value of "noEmitOnError"
The docs on "Using TS Effectively" say "noEmitOnError" default is `true` (and suggest leaving this value as-is): https://github.com/typed-ember/ember-cli-typescript/blame/master/docs/ts/using-ts-effectively.md#L30

It seems the actual value in the blueprint is `false`, though: https://github.com/typed-ember/ember-cli-typescript/blob/master/blueprint-files/ember-cli-typescript/tsconfig.json#L16

Seems likely this docs line should be replaced with an explanation of why the value is as it is, and when to change it (or, if it really is a good idea to have it enabled, perhaps the blueprint value should be changed?)

MrChocolatine:
Fix the value of the tsconfig option `noEmitOnError` to `true`, to be consistent with what is currently mentioned in the documentation.

Fix #1402 .

### DIFF
--- a/blueprint-files/ember-cli-typescript/tsconfig.json
+++ b/blueprint-files/ember-cli-typescript/tsconfig.json
@@ -13,7 +13,7 @@
     "noUnusedLocals": true,
     "noUnusedParameters": true,
     "noImplicitReturns": true,
-    "noEmitOnError": false,
+    "noEmitOnError": true,
     "noEmit": true,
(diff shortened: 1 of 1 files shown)


dfreeman:
Thank you, @MrChocolatine!

PR #1424, opened 2021-04-21
fix: implement `cacheKeyForTree` that returns `null`

brendenpalmer:
fix: implement `cacheKeyForTree` that returns `null`

This fixes a bug between `ember-cli-typescript`, `ember-engines` dedupe, and `ember-cli-babel`. `ember-cli-babel` uses the existence of `ember-cli-typescript` as a dependency to determine whether to process `*.ts` files; however, `ember-engines` dedupe is _stateful_ so it's possible for `ember-cli-typescript` to not be a dependency when `ember-cli-babel` is running.

For more info on `ember-engines` dedupe logic: https://github.com/ember-engines/ember-engines/blob/master/packages/ember-engines/lib/utils/deeply-non-duplicated-addon.js#L35

For more info on how `ember-cli-babel` determines whether to process `*.ts`: https://github.com/babel/ember-cli-babel/blob/master/lib/babel-options-util.js#L407

### DIFF
--- a/ts/addon.ts
+++ b/ts/addon.ts
@@ -48,6 +48,25 @@ export default addon({
     return `${__dirname}/blueprints`;
   },
 
+  /**
+   * We implement a custom `cacheKeyForTree` here to prevent `ember-engines`
+   * from deduping this addon. This fixes a bug between `ember-cli-typescript`,
+   * `ember-engines` dedupe, and `ember-cli-babel`. `ember-cli-babel` uses the
+   * existence of `ember-cli-typescript` as a dependency to determine whether
+   * to process `*.ts` files; however, `ember-engines` dedupe is _stateful_
+   * so it's possible for `ember-cli-typescript` to not be a dependency when
+   * `ember-cli-babel` is running.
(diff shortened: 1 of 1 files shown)


dfreeman (approved):
This seems reasonable to me! It's a bit of a shame we need to invalidate the caching here since it would otherwise be totally valid, but since we don't _have_ any trees of runtime code the impact should be pretty minimal.

brendenpalmer:
@dfreeman I agree. One of the alternatives is to update `ember-cli-babel` to _always_ transpile `*.ts` files, which is something I'm exploring in conjunction w/ this. If a fix in `ember-cli-babel` can land sooner than I expect, then I'll revert this PR (if it's merged by then), or close this out.

jamescdavis:
@brendenpalmer https://github.com/babel/ember-cli-babel/pull/400 resolved this, right?

brendenpalmer:
@jamescdavis Yes, this has been resolved in `ember-cli-babel` itself, closing this PR out

PR #1461, opened 2022-01-04
WIP: support composite projects

### LINKED ISSUE #234: Does not work with monorepo / yarn workspaces
I do have a setup with a monorepo / yarn workspace and lerna. Mostly identical with `ember-decorators` repo. For reference I added `lerna.json` and `package.json` (for the workspace) as reference below. The structure typically looks like this:

```
[root]
|- packages
|-- packageA
|-- packageB
|- node_modules
|- package.json
`- lerna.json
```

So, when `packageA` requires `ember-cli-typescript` it is installed to `[root]/node_modules/ember-cli-typescript` and so is `typescript`. Hence the watcher is setup to watch in the folder `[root]/packages/packageA/node_modules/***` which doesn't exist. 

As such the watcher tries to endlessly find something that doesn't exist. There should be some mechanism first to detect, wether this is a yarn workspace and an which folders to look for (it still can be the package related `node_modules` folder).

```json
// lerna.json
{
  "lerna": "2.11.0",
  "npmClient": "yarn",
  "useWorkspaces": true,
  "packages": [
    "packages/*"
  ],
  "version": "0.0.0"
}
```

```json
// package.json
{
	"private": true,
	"devDependencies": {
		"lerna": "^2.11.0"
	},
	"workspaces": [
		"packages/*"
	]
}
```

### LINKED ISSUE #1451: Type checking failure when building in-repo addons with project references
```
ember-cli: 3.28.0
node: 14.15.1
os: darwin x64
```

### Please paste the output of `tsc -v` here
```
Version 4.4.2
```

### Please paste the version of `ember-cli-typescript` and `ember-cli-typescript-blueprints` here
```
ember-cli-typescript@4.2.1
```
**no blueprints**

### Please paste your `tsconfig.json` and `tslint.json` or `eslint.json` (if applicable) below
[top level tsconfig](https://github.com/brendenpalmer/ember-cli-typescript-typechecking-repro/blob/master/tsconfig.json)
[nested tsconfig](https://github.com/brendenpalmer/ember-cli-typescript-typechecking-repro/blob/master/packages/ts-addon/tsconfig.json)

```sh
git clone git@github.com:brendenpalmer/ember-cli-typescript-typechecking-repro.git
cd ember-cli-typescript-typechecking-repro
yarn
cd packages/ts-addon/
ember b
```

typechecking should fail with the above steps, running ember build from inside ts-addon. This repo make use of composite projects & in this particular case, the ts compiler fails to build the TS files in ts-addon-b, thus throwing a typecheck error.

##### Reproduction Case
**Link: ** https://github.com/brendenpalmer/ember-cli-typescript-typechecking-repro

Instructions listed above.

I expected to successfully run `ember b` in the sample app's `ts-addon` package without type errors

### What happened instead?
Typechecking fails with the following error:
```
error TS6305: Output file '/Users/hmajoros/Job/ember-cli-typescript-typechecking-repro/packages/ts-addon-b/addon/index.d.ts' has not been built from source file '/Users/hmajoros/Job/ember-cli-typescript-typechecking-repro/packages/ts-addon-b/addon/index.ts'.

1 import { FOO_BAR_BAZ } from 'ts-addon-b';
```

chriskrycho:
Support composite projects and project references to do incremental builds.

- [ ] hook into the basic TS config for the build mode
- [ ] make sure that things actually work in workspaces correctly
- [ ] add tests!

This resolves some long-standing gaps:

- Resolves #234
- Resolves #1451

### DIFF
--- a/ts/lib/typechecking/worker/index.ts
+++ b/ts/lib/typechecking/worker/index.ts
@@ -8,6 +8,8 @@ import {
   CompilerOptions,
   WatchOfConfigFile,
   WatchCompilerHostOfConfigFile,
+  SolutionBuilderWithWatchHost,
+  SolutionBuilder,
 } from 'typescript';
 
 const debug = logger('ember-cli-typescript:typecheck-worker');
@@ -31,10 +33,14 @@ export default class TypecheckWorker {
   private lastSettledStatus = this.status;
   private typecheckTimeout?: NodeJS.Timer;
 
-  private projectRoot!: string;
-  private ts!: typeof import('typescript');
-  private watch!: WatchOfConfigFile<SemanticDiagnosticsBuilderProgram>;
-  private compilerOptions!: CompilerOptions;
+  // SAFETY: these are configured during the `start` call, which in turn is run
+  // after construction but *before* anything else happens.
+  private declare projectRoot: string;
+  private declare ts: typeof import('typescript');
+  private declare watch:
+    | WatchOfConfigFile<SemanticDiagnosticsBuilderProgram>
+    | SolutionBuilder<SemanticDiagnosticsBuilderProgram>;
+  private declare compilerOptions: CompilerOptions;
 
   /**
    * Begin project typechecking, loading TypeScript from the given project root.
@@ -42,6 +48,13 @@ export default class TypecheckWorker {
   public start(projectRoot: string) {
     this.projectRoot = projectRoot;
     this.ts = this.loadTypeScript();
+
+    // In the case of a composite project
+    if (this.compilerOptions.composite) {
+      this.watch = this.ts.createSolutionBuilderWithWatch(this.buildCompositeHost(), [], {}, {});
+      return;
+    }
+
     this.watch = this.ts.createWatchProgram(this.buildWatchHost());
     this.compilerOptions = this.watch.getProgram().getCompilerOptions();
   }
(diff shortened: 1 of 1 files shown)


ts/lib/typechecking/worker/index.ts
```diff
@@ -31,17 +33,28 @@ export default class TypecheckWorker {
...
 
   /**
    * Begin project typechecking, loading TypeScript from the given project root.
    */
   public start(projectRoot: string) {
     this.projectRoot = projectRoot;
     this.ts = this.loadTypeScript();
+
+    // In the case of a composite project
+    if (this.compilerOptions.composite) {
```
chriskrycho, line 53: Note: here we don't actually *have* a `this.compilerOptions`. `declare` and `!` alike mislead here.

However, we currently only use `compilerOptions` as a way of handling the `noEmitOnError` setting, which [seems not to be used in composite projects anyway](https://discord.com/channels/@me/689455032629854295/927735143093919775). For that setting, we can refactor the design here to simply match the `tsc` behavior (and document it). That just means making `compilerOptions` optional.

However, as this line itself indicates, we do actually need to load and resolve the project `tsconfig.json` to even know which mode to build in, so I will follow up by introducing *that* here.

This also leads to an obvious question, though: should we be using `noEmitOnError`, which has caused a great deal of back-and-forth and confusion about exactly how it is integrated into the build? Or should we allow users to opt into `ember-cli-typescript`-specific configuration in some way, e.g. via something like an `ember.typescript` settings object in `package.json`?

vitch:
Could you elaborate a little more on what this will achieve once it's done?

I think it's related to the thing I'm running into over [here](https://github.com/vitch/ts-interface/issues/1) except that I'm not even sure I'm running this code over there!

The issue I'm seeing over there happens when running `tsc --build --watch` directly (and `ember s` works fine). 

In the actual monorepo I'm reproducing we don't currently use `ember-cli-typescript` and instead just rely on `enableTypeScriptTransform` from `ember-cli-babel` to strip types when building (and "just typescript" for editor support and type checking on CI)...

chriskrycho:
@vitch sorry, I totally missed your comment a few months ago! The short version is is: `ember-cli-typescript` currently doesn't understand project references *at all*, and so doesn’t support building the intermediate `.d.ts` and `.tsbuildinfo` files, and doesn't work if you are working in a package within a workspace.

The thing you hit with interface merging appears to me to be something different (though I haven't poked at it in detail)—it is, in fact, the *desired* behavior for interface merging in that you don’t *want* interface declarations from non-referenced projects to be available. Without that, for example, there would be no way to have any kind of isolation between packages in a workspace; they would all automatically share all merged interfaces. What you want (and what composite projects give you!) is an actual explicit dependency graph: you will only inherit the merged interfaces from the projects you depend on.

vitch:
@chriskrycho - thanks for the response...

I think I got to the bottom of the interface merging thing and it was that the interfaces are only merged from files which are explicitly imported or are `include`d from a `tsconfig`. IIRC I thought I was seeing different behavior with the registry for Ember services but in the end it was behaving the same.

I actually put this whole thing on a back burner for a while because there was a bunch of other stuff that we need to update before we can even benefit from what I was trying to achieve. In the meantime we might be able to make the relevant files v2 addons and thus avoid the mass of composite projects which are the cause of many of our current problems anyway :)

chriskrycho:
Closing this: per discussion (and forthcoming RFC from the Ember TypeScript team!) we are *not* going to implement this, and instead are going to shift how `ember-cli-typescript` works with the rest of the ecosystem rather substantially. More details to come. (Glint `--build` support *is* still coming, though!)

PR #1462, opened 2022-01-08
Blueprints: improve specificity of `locationType` in config

chriskrycho:
The only legal types here are `'history'` and `'hash'`. Specify as much in the type we generate.


chriskrycho:
Closing in favor of #1466, which is more correct.

PR #1466, opened 2022-01-23
Fix type of `locationType`

NullVoxPopuli:
In fresh apps, previously, we'd start with a TS error, which crashes ember-cli atm, due to an ember-cli issue.


chriskrycho (approved):
👍🏼

PR #1467, opened 2022-01-24
Add more specific `locationType` type to match new ember types

mydea:
With the new ember 4 types (https://github.com/DefinitelyTyped/DefinitelyTyped/pull/58060), the `locationType` expects a more specific type than "string".


mydea:
Oops, sorry, totally missed that! :see_no_evil:

PR #1468, opened 2022-01-25
Update types after Ember v4 types release

chriskrycho:
Correctly aligning the types in DefinitelyTyped meant these types were in turn out of sync, and needed to be targeted correctly.

This *should* get CI happy again. 🤞🏼


chriskrycho:
Superseded by #1472.

Issue #1470, opened 2022-01-25
Types for ember-test-helpers are confusing


mydea:
### Which package(s) does this problem pertain to?
  - [ ] @types/ember
  - [ ] @types/ember__string
  - [ ] @types/ember__polyfills
  - [ ] @types/ember__object
  - [ ] @types/ember__utils
  - [ ] @types/ember__array
  - [ ] @types/ember__engine
  - [ ] @types/ember__debug
  - [ ] @types/ember__runloop
  - [ ] @types/ember__error
  - [ ] @types/ember__controller
  - [ ] @types/ember__component
  - [ ] @types/ember__routing
  - [ ] @types/ember__application
  - [ ] @types/ember__test
  - [x] @types/ember__test-helpers
  - [ ] @types/ember__service
  - [ ] @types/ember-data
  - [ ] @types/rsvp
  - [ ] Other
  - [ ] I don't know

The whole situation around types for @ember/test-helpers is rather confusing to me.
There is `@types/ember__test-helpers`, as well as `@types/ember-test-helpers`.

This leads to some confusion. Also, especially with the v4 type update, `@types/ember-test-helpers` seems to not have been updated anymore (it still relies on `@types/ember@^3`), which leads to a bunch of issues with conflicting types for me.

At the end, I could resolve it by removing `@types/ember-test-helpers` and just using `@types/ember__test-helpers@^4`. This also means you have to import everything from `@ember/test-helpers` - previously, there have been some things like the TestContext that you used to have to import from `ember-test-helpers` instead. This is also still that way in the docs.

IMHO it probably makes sense to remove all mentions and traces of `ember-test-helpers` and `@types/ember-test-helpers`

chriskrycho:
Your recommendation is correct. (The ecosystem was confusing! Our types just had to try to represent it. 😩)

- You should remove `ember-test-helpers` and corresponding types from your app entirely, switching entirely to `@ember/test-helpers` and `@types/ember__test-helpers`.
- We need to update the docs to stop referencing `ember-test-helpers`.

PR #1472, opened 2022-01-25
Update types & ember to 4.x

mydea:
This PR updates all dev dependencies of types to their latest version, as far as possible.
It also updates the documentation for how to import/extend the `TestContext` type.

Closes https://github.com/typed-ember/ember-cli-typescript/issues/1470


chriskrycho:
Thanks for this! I have been independently working on a similar branch, so this is great. I think it needs one more change: bumping the supported Node version everywhere it's specified (`engines`, CI, etc.) up to at least v12 – that will make the remaining CI jobs pass, I think.

mydea:
I have updated node to 12 everywhere (and setup CI to use the pinned volta version, so this only needs to be set in a single place).

FYI I have also tried to update all other dev dependencies (basically `yarn ember-cli-update` etc.), but that lead to major issues with the tests. I think it has something to do with how ember-auto-import etc. works, and probably requires some more thought - e.g. the skeleton app for the build testing was not really outputting the given `app/app.ts` file at all, for whatever reason, but the generated `dist/asets/skeleton-app.js` file had a content like this:

```js
'use strict';



;

;define('skeleton-app/config/environment', [], function() {
  var prefix = 'skeleton-app';
try {
  var metaName = prefix + '/config/environment';
  var rawConfig = document.querySelector('meta[name="' + metaName + '"]').getAttribute('content');
  var config = JSON.parse(decodeURIComponent(rawConfig));

  var exports = { 'default': config };

  Object.defineProperty(exports, '__esModule', { value: true });

  return exports;
}
catch(err) {
  throw new Error('Could not read config from meta tag with name "' + metaName + '".');
}

});

;
          if (!runningTests) {
            require("skeleton-app/app")["default"].create({});
          }
        
//# sourceMappingURL=skeleton-app.map
```

If there is anybody is interesting in looking at what I've done so far, you can see it here:
https://github.com/mydea/ember-cli-typescript/tree/fn/update-ember-cli

PR #1485, opened 2022-02-22
Encourage the use of type-only imports

muziejus:
When I first started setting up my models, I was getting `import/no-cycle` errors. I've now added some language to encourage the used of `import type` in the model files, and I've also changed the examples to indicate that usage.

### DIFF
--- a/docs/ember-data/models.md
+++ b/docs/ember-data/models.md
@@ -62,7 +62,11 @@ export default class User extends Model {
 
 Relationships between models in Ember Data rely on importing the related models, like `import User from './user';`. This, naturally, can cause a recursive loop, as `/app/models/post.ts` imports `User` from `/app/models/user.ts`, and `/app/models/user.ts` imports `Post` from `/app/models/post.ts`. Recursive importing triggers an [`import/no-cycle`](https://github.com/import-js/eslint-plugin-import/blob/main/docs/rules/no-cycle.md) error from eslint. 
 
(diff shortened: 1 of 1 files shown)


docs/ember-data/models.md
```diff
@@ -58,7 +58,13 @@ export default class User extends Model {
 }
 ```
 
-## `@belongsTo`
+## Relationships
+
+Relationships between models in Ember Data rely on importing the related models, like `import User from './user';`. This, naturally, can cause a recursive loop, as `/app/models/post.ts` imports `User` from `/app/models/user.ts`, and `/app/models/user.ts` imports `Post` from `/app/models/post.ts`. Recursive importing triggers an [`import/no-cycle`](https://github.com/import-js/eslint-plugin-import/blob/main/docs/rules/no-cycle.md) error from eslint. 
+
+One should be certain to make use of [type-only imports](https://www.typescriptlang.org/docs/handbook/release-notes/typescript-3-8.html), available since TypeScript 3.8, and write `import type User from './user';`.
```
chriskrycho, line 65:
```suggestion
To avoid these errors, use of [type-only imports](https://www.typescriptlang.org/docs/handbook/release-notes/typescript-3-8.html), available since TypeScript 3.8:

```ts
import type User from './user';
```
```

chriskrycho (approved):
Thank you, this is a great little improvement!

PR #1488, opened 2022-03-03
Update tsconfig and corresponding docs

chriskrycho:
Our generated blueprint was *years* out of date, and between the changes we made in the types over the last year or so and the guidance we have given in the TS RFCs we are now both *able* and *required* to set `strict: true` as well as to opt into `noUncheckedIndexedAccess` and to disable `allowSyntheticDefaultImports` and `esModuleInterop`.

Additionally, this provided a nice opportunity to document the choices we have made so folks have something to latch onto when looking at the config we generate for them, and to update the config to target the current latest versions of the `target` and `module` fields.


blueprint-files/ember-cli-typescript/tsconfig.json
```diff
@@ -1,25 +1,67 @@
...
+    // Interop: these are viral and will require anyone downstream of your
+    // package to *also* set them to true. If you *must* enable them to consume
+    // an upstream package, you should
+    "allowSyntheticDefaultImports": false,
+    "esModuleInterop": false,
+
+    // --- Lint-style rules
+
+    // You should feel free to change these, especially if you are already
+    // covering them via linting (e.g. with @typescript-eslint).
```
NullVoxPopuli, line 30: on _very large_ projects, would disabling these improve build speed?

blueprint-files/ember-cli-typescript/tsconfig.json
```diff
@@ -1,25 +1,67 @@
...
     "noUnusedLocals": true,
     "noUnusedParameters": true,
     "noImplicitReturns": true,
+    "noPropertyAccessFromIndexSignature": true,
+
+    // --- Compilation/integration settings
+    // Setting `noEmitOnError` here allows ember-cli-typescript to catch errors
+    // and inject them into Ember CLI's build error reporting, which provides
+    // nice feedback for when
     "noEmitOnError": true,
```
NullVoxPopuli, line 41: obligatory: https://github.com/ember-cli/ember-cli/issues/9584 (for reference, in case people are investigating the issue)

I _have_ to flip noEmitOnError to false in every app, otherwise the dev server crashes.
(and I haven't had time to fix it apologies, everyone)

NullVoxPopuli (approved):
Thanks for adding all the context/comments! <3

blueprint-files/ember-cli-typescript/tsconfig.json
```diff
@@ -1,25 +1,67 @@
...
+    // Ember makes heavy use of decorators; TS does not support them at all
+    // without this flag.
+    "experimentalDecorators": true,
+
+    // Support generation of source maps. Note: you must *also* enable source
+    // maps in your `ember-cli-babel` config and/or `babel.config.js`.
+    "declaration": true,
+    "declarationMap": true,
     "inlineSourceMap": true,
     "inlineSources": true,
```
dfreeman, line 59: Given that we set `noEmit: true` here and use Babel for actually emitting code, do these flags have any effect?

blueprint-files/ember-cli-typescript/tsconfig.json
```diff
@@ -1,25 +1,67 @@
...
+    // Strictness settings -- you should *not* change these: Ember code is not
+    // guaranteed to type check with these set to looser values.
+    "strict": true,
+    "noUncheckedIndexedAccess": true,
+
+    // Interop: these are viral and will require anyone downstream of your
+    // package to *also* set them to true. If you *must* enable them to consume
+    // an upstream package, you should
+    "allowSyntheticDefaultImports": false,
+    "esModuleInterop": false,
```
dfreeman, line 25: In theory it's fine for apps to enable these flags, isn't it? Since they _have_ no downstream consumers, and they know what packager they're using and will quickly discover if it turns out not to have ES interop enabled?

docs/ts/using-ts-effectively.md
```diff
@@ -8,24 +8,28 @@ Some specific tips for success on the technical front:
...
+    "noUnusedLocals": true,
+    "noUnusedParameters": true,
+    "noImplicitReturns": true,
+    "noPropertyAccessFromIndexSignature": true,
+  }
 }
 ```
 
-A good approach is to start at your "leaf" files (the ones that don't import anything else from your app, only Ember types) and then work your way back inward toward the most core types that are used everywhere. Often the highest-value modules are your Ember Data models and any core services that are used everywhere else in the app – and those are also the ones that tend to have the most cascading effects (having to update _tons_ of other places in your app) when you type them later in the process.
+A good approach is to start at your "leaf" modules (the ones that don't import anything else from your app, only Ember or third-party types) and then work your way back inward toward the most core moduels that are used everywhere. Often the highest-value modules are your Ember Data models and any core services that are used everywhere else in the app – and those are also the ones that tend to have the most cascading effects (having to update _tons_ of other places in your app) when you type them later in the process.
```
dfreeman, line 32:
```suggestion
A good approach is to start at your "leaf" modules (the ones that don't import anything else from your app, only Ember or third-party types) and then work your way back inward toward the most core modules that are used everywhere. Often the highest-value modules are your Ember Data models and any core services that are used everywhere else in the app – and those are also the ones that tend to have the most cascading effects (having to update _tons_ of other places in your app) when you type them later in the process.
```

blueprint-files/ember-cli-typescript/tsconfig.json
```diff
@@ -1,25 +1,67 @@
 {
   "compilerOptions": {
-    "target": "es2020",
-    "allowJs": true,
+    "target": "ES2022",
+    "module": "ES2022",
     "moduleResolution": "node",
```
dfreeman, line 5: Not actionable yet, but I suspect we're going to want to change this to `node12` as soon as that graduates from nightly, since Webpack 5 (and so by extension, `ember-auto-import` and Embroider) honor `exports` config.

blueprint-files/ember-cli-typescript/tsconfig.json
dfreeman: I think the only way disabling these would have an impact on build/typecheck speed would be if you had an absolutely _massive_ number of violations, in which case the time spent printing those out could add up. The information they act on, though, is already being gathered as part of typechecking, so I wouldn't expect enabling them or disabling them to otherwise make much of a difference.

dfreeman (commented):
🎉  Thank you!

blueprint-files/ember-cli-typescript/tsconfig.json
chriskrycho: I think that's correct. I'll update the comment to that effect.

blueprint-files/ember-cli-typescript/tsconfig.json
chriskrycho: Yep. Hoping they ship that in 4.7 and we can flip it on. 🤞🏼

PR #1489, opened 2022-03-03
typo: comment should be pluralized.

muziejus:
```js
import Model, { hasMany, type AsyncHasMany, type SyncHasMany } from '@ember-data/model';
import type Comment from './comment';
import type User from './user';

export default class Thread extends Model {
  @hasMany('comment')
  declare comment: AsyncHasMany<Comment>;

  @hasMany('user', { async: false })
  declare participants: SyncHasMany<User>;
}
```

Pretty sure `comment` should be `comments`.


chriskrycho (approved):
Yep! Thank you!

PR #1490, opened 2022-03-04
docs: fix typos

muziejus:
Currently in the docs, `@attr() declare name?` is typed as `string`, and no example of
`@attr('string')` is given. I'm not sure if showing in the documentation
that not passing a transform to `@attr()` should explicitly default to
string is the correct thing to do, but I definitely think we should
model the use of `@attr('string')` in the code examples.

### DIFF
--- a/docs/ember-data/models.md
+++ b/docs/ember-data/models.md
@@ -23,7 +23,7 @@ import Model, { attr } from '@ember-data/model';
 import CustomType from '../transforms/custom-transform';
 
 export default class User extends Model {
-  @attr('string')
+  @attr()
   declare name?:  string;
 
   @attr('number')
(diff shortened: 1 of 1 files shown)


docs/ember-data/models.md
```diff
@@ -23,8 +23,8 @@ import Model, { attr } from '@ember-data/model';
 import CustomType from '../transforms/custom-transform';
 
 export default class User extends Model {
-  @attr()
-  name?:  string;
+  @attr('string')
+  declare name?:  string;
```
chriskrycho, line 27: Let's split this one and the one like it below into a separate PR: that way we can land the typo fixes while we think about/chat with Data folks about this part.

chriskrycho (changes requested):
Good catches. I'd like to hear from Data folks on the `@attr()` default, so once you split that out, let me know and I'll merge the typo fixes. Thanks!

docs/ember-data/models.md
muziejus: Done.

chriskrycho (approved):

chriskrycho:
Thanks!

PR #1504, opened 2022-05-05
docs: fix spelling resolveable -> resolvable

jacobq:
These links lead to pages that say:
"The word you've entered isn't in the dictionary."
https://www.merriam-webster.com/dictionary/unresolveable
https://www.merriam-webster.com/dictionary/resolveable

I guess we who are picky about types can also be sticklers about spelling, or maybe it's something like OCD about those "squiggles" in the IDE....


chriskrycho:
Thank you!

PR #1509, opened 2022-05-24
Remove blueprints package

NullVoxPopuli:
The blueprints package causes confusion as it reintroduces a bunch of concepts that are not present in modern apps. And no blueprints are better than incorrect blueprints.


wagenet:
This would also resolve the issue https://rfcs.emberjs.com/id/0776-typescript-blueprints/.

chriskrycho:
Sorry for the delay, we agree this is the right path here. CI is likely to be unhappy for unrelated reasons, and I'm going to see if I can get that unblocked and then will update this again if need be!

chriskrycho:
Should be unblocked by #1531; I will do the update-branch-and-merge dance once that one lands.

PR #1511, opened 2022-05-25
docs: Overhaul component documentation to use Signatures

muziejus:
The current documentation on components only discusses the arguments
passed to a Glimmer component. Given RFC 0748's implementation of a
`Signature` type with `Element`, `Args`, and `Blocks` fields, I think
it's appropriate to update the component docs accordingly.

The earlier version of the page spent a lot of time discussing the
minutiae of how `this.args` comes to be typed, etc., and focused on the
component's `constructor`. Both seem to me a little off-track for someone
eager to write an Octane app in TypeScript, especially given how native
class syntax intercepts the need for a constructor.

I've instead broken up the document into a section with no signature,
the signature itself, args, blocks, element, and then, finally, how to
subclass one's own component.

A lot of the text I've adapted/copied from either RFC 0748 or written
based on advice (esp wrt subclassing) I've gotten from @chriskrycho on Discord.
I think it's a good refresh of documentation that is a few years old.
It's also looking forward, I hope, to landing some of this documentation
in Ember proper.

If the discussion of constructors and `owner` is still important (I'm
basing this on my personal experience), then we can add it back in, but
I'd need help updating that section, I think.

### DIFF
--- a/docs/ember/components.md
+++ b/docs/ember/components.md
@@ -4,19 +4,21 @@
 New to Ember or the Octane edition specifically? You may want to read [the Ember Guides’ material on `Component`s](https://guides.emberjs.com/release/components/) first!
 {% endhint %}
 
-Glimmer Components are defined in one of three ways: with templates only, with a template and a backing class, or with only a backing class \(i.e. a `yield`-only component\). When using a backing class, you get a first-class experience using TypeScript! Unfortunately, we don’t yet support type-checking for templates, but we hope to build that out eventually. Don’t let that stop you, though: types in your component classes make for a great experience, so let’s dig in and see how it works in practice.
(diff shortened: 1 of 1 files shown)


chriskrycho (changes requested):
Hey @muziejus – thanks for the effort you put in here. I agree that we want to emphasize signatures and the Octane happy path (a lot of this does indeed need a reshuffle!), but I am not entirely sure about the *specific* direction here. I'm happy to work with you on it to get it where we want, including hopping on a call, though.

Two overall notes here, which will inform further discussion we should have about how to rework this stuff:

1. Please avoid making stylistic changes to the prose or especially things like changing the headings here. We gave the structure quite a bit of thought when we wrote it, so changing those things is not something we want to just tackle in a PR.

2. As a general guideline—rather than starting with a PR, start with opening an issue, so that we can have a conversation about these kinds of things *before* you put in the work! We hate ending up in a situation where you might feel like you wasted your time. 😩

chriskrycho:
Thanks again for the first pass here, @muziejus! I'm going to close this at this point, as I am planning to (later this month or in June) do a full pass with pulls together *all* of the TypeScript materials into a single coherent story that can be a good foundation for future integration into Ember’s own docs, etc.

PR #1517, opened 2022-07-20
Update testing.md

ChristopheTiet:
import hbs from 'ember-cli-htmlbars-inline-precompile' is deprecated
instead use import { hbs } from 'ember-cli-htmlbars'


chriskrycho (approved):
Thank you!

PR #1526, opened 2022-09-24
Include `@types/ember__destroyable` when running the default blueprint

### LINKED ISSUE #1497: Newly installed app doesn't get @types/ember__destroyable
### Which package(s) does this problem pertain to?
  - [x] Other

(Your bug form doesn't include @types/ember__destroyable)

Create a new Ember app with ember 4.1, add ember-cli-typescript including running its blueprint.

Try to import from `@ember/destroyable` in your app.  It's a type error.

You can manually install `@types/ember__destroyable` to solve the problem.

bertdeblock:
Closes #1497.


chriskrycho (approved):
💙

chriskrycho:
Not: I expect this may fail CI because of smoke tests. If so, this fix should be straightforward: just update the fixtures.

chriskrycho:
I've updated this with the changes from `master`, and it passes locally, so I've pushed that to this branch and it should go green on everything but the ember-cli checks. Once that's true, I'll merge this!

PR #1196, opened 2020-07-02
[v4] Use prepack and postpack

jamescdavis:
Use `prepack` and `postpack` instead of `prepublishOnly` and `postpublish`.

As of https://github.com/yarnpkg/yarn/pull/5712 `yarn` runs `prepack` and `postpack` when publishing (`npm` always did).

PR #1527, opened 2022-09-24
docs: use `prepack` and `postpack` instead of `prepublishOnly` and `postpublish`

bertdeblock:
This was changed in https://github.com/typed-ember/ember-cli-typescript/pull/1196.


chriskrycho (approved):
Ah, thank you!

PR #1528, opened 2022-09-28
Update `@types/ember-qunit` and `@types/qunit`

bertdeblock:
Attempt at resolving the CI failures over in #1526 and #1527.


chriskrycho:
This is apt to break *much* harder, since it has introduced a second version of `@types/ember`; you'll need to dedupe those.

chriskrycho:
Looks like we need to make sure *all* the `@types/ember*` packages get bumped.

chriskrycho:
Should be unblocked by #1531; I will do the update-branch-and-merge dance once that one lands.

PR #1533, opened 2022-10-25
Update to use Node 14+ (drop support for Node 12)

chriskrycho:
- in `volta` key in `package.json`
- in config for Windows CI in `ci.yml`


chriskrycho:
All right, landing this because it lets proceed with Ember CLI fixes (of which there is actually one, likely because of a transitive dependency change or some such, but it was made invisible by the Node 14 issues previously).

PR #1534, opened 2022-10-25
Use volta action v4 for all targets

chriskrycho:
- Update to v4 for all existing targets
- Switch to using the Volta action for Windows targets


chriskrycho:
As with #1533, this is unrelated to the Ember CLI failures, and it nicely improves our CI, so here we go!

PR #1536, opened 2022-10-25
Set up release-it for publishing

chriskrycho:
I never want to do the chore I did this morning for 5.2.0 again! 😂


chriskrycho:
The last thing here is to tweak the `mocha` invocation for *CI* to work (it works in development using `ts-node` but not in regular JS tests 🙄).

dfreeman:
Looks good!

PR #1537, opened 2022-10-26
Airlift `updatePathsForAddon` from the blueprints package

dfreeman:
It turns out the default blueprint relied on the fact that we were adding `ember-cli-typescript-blueprints` to import an `updatePathsForAddon` utility function. Since that's no longer a safe assumption, I've just airlifted that function directly into this codebase.

I also updated a couple places in the docs that assumed the presence of `ember-cli-typescript-blueprints` by default.

### DIFF
--- a/package.json
+++ b/package.json
@@ -121,7 +121,8 @@
     "typescript": "4.5.5"
   },
   "resolutions": {
-    "hawk": "7"
+    "hawk": "7",
+    "ember-cli-typescript": "link:."
   },
   "engines": {
     "node": ">= 12.*"
--- a/package.json
+++ b/package.json
@@ -89,7 +89,6 @@
     "ember-cli-htmlbars": "5.3.1",
     "ember-cli-inject-live-reload": "2.0.2",
     "ember-cli-sri": "2.1.1",
-    "ember-cli-typescript-blueprints": "3.0.0",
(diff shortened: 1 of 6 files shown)


package.json
```diff
@@ -122,7 +121,8 @@
     "typescript": "4.5.5"
   },
   "resolutions": {
-    "hawk": "7"
+    "hawk": "7",
+    "ember-cli-typescript": "link:."
```
chriskrycho, line 125: Huh, is this a standard Yarn pattern for self-resolution? I hadn't seen it before.

On reading the lock file changes, how does this impact consumers (if at all)?

chriskrycho (approved):

package.json
dfreeman: > Huh, is this a standard Yarn pattern for self-resolution?

I'm not sure if it's "standard" but it's definitely one I've used in a few places. In principle it could blow up on us in a future where we make big breaking changes here and our dependencies that rely on `ember-cli-typescript` aren't ready for that but are having it forced on them anyway. Since I don't think we're super likely to be making any changes like that, though... ¯\\\_(ツ)_/¯ 

> On reading the lock file changes, how does this impact consumers (if at all)?

I don't think it should impact consumers at all, since `resolutions` only apply in your project's own `package.json`, not your dependencies'.

PR #1545, opened 2023-01-12
Remove deprecated `@types/ember__test-helpers` from blueprint

simonihmig:
The type package has been deprecated and is basically a no-op package, with types coming directly from `@ember/test-helpers`, so can be removed here.


simonihmig:
CI is failing for floatings deps, but same happens with same errors for `master`: https://github.com/typed-ember/ember-cli-typescript/actions/runs/3865785545

chriskrycho:
Thank you!

PR #1554, opened 2023-05-09
docs: Reference Glint in docs

IgnaceMaes:
This PR updates the docs to refer to Glint for template type checking.

For now the most obvious case has been adjusted, but a full review would be good to verify if there are more occurrences.


chriskrycho (approved):
This is a good update; thank you!

PR #1562, opened 2023-06-30
Adding note to testing.md about qunit-dom

tsengia:
This PR just adds a small snippet to the `testing.md` page about importing `qunit-dom` so that Typescript is aware of the `assert.dom()` function commonly used in Component tests.

It took me a while to figure this out, so I hope that this small snippet can save others some time in the future.


chriskrycho (approved):
Ah, this is good to have here. qunit-dom's own docs *also* cover this, and I believe the blueprints shipped in `ember-source` do the right thing, but this is great to have. Thanks!

PR #1570, opened 2023-09-28
Use the current Node LTS

chriskrycho:
Note: This _is_ a breaking change! However, it is only breaking in a strict sense, and no runtime behavior should be different.


NullVoxPopuli:
We  have a bit of a catch-22 -- the floating dependencies tests are not capable of passing until this is merged. But this typically can't merge until CI passes. So, we're going to have red c.i. for a smidge. Going to merge after changing the node version to 18.18.1 from 18.18.0, as .0 had a bug that affected ember projects.

NullVoxPopuli:
hm, merge queue isn't going to work here

PR #1575, opened 2023-10-12
re-roll lockfile, fixing the floating deps tests

NullVoxPopuli:
Floating tests have been failing, but in order to fix them, some dependency changes needed to occur.

This started in a few other PRs, but I found that this was the minimal changeset I could generate while still getting things working not-worse than `main` today.


NullVoxPopuli:
Just kidding, looks like this _has_ to be rolled up in to: https://github.com/typed-ember/ember-cli-typescript/pull/1574

